PR #226, opened 2015-01-14
Add posibility to define model_dir as an array of directorie paths

capripot:
Sometimes in projects you have shared modules that contains shared models.
This patch adds the possibility to define multiple dirs in the option --model-dir.


ctran:
Thanks for the PR.  Would you mind cleaning up the commits by squashing/rebase them into 1 or 2?

capripot:
I just cleaned my commits and corrected some bugs as well.

PR #227, opened 2015-01-14
use setup_options instead of set_defaults in rake task template

capripot:
The template for automated rake task should probably use `setup_options` method instead of `set_defaults`


capripot:
Its actually a mistake, I should have patched lib/tasks/annotate_models.rake#21 instead. I did it in pull request #226

PR #231, opened 2015-02-19
Prefer SQL column type over normalized AR type

vassilevsky:
We have columns of type `point` in our PostgreSQL database:

```
#<ActiveRecord::ConnectionAdapters::PostgreSQLColumn:0x007ffd7b829aa8
 @coder=nil,
 @default=nil,
 @limit=nil,
 @name="point",
 @null=false,
 @precision=nil,
 @primary=false,
 @scale=nil,
 @sql_type="point",
 @type=:string>
```

http://www.postgresql.org/docs/9.4/interactive/datatype-geometric.html

They are annotated as `string` in our models. This is misleading: `point` is actually 2 floats; not even close to a string.

I thought that SQL type should take precedence.

I understand that this change may be not enough or is completely wrong, and would welcome suggestions on the best solution.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -140,7 +140,7 @@ module AnnotateModels
         attrs << "not null" unless col.null
         attrs << "primary key" if klass.primary_key && (klass.primary_key.is_a?(Array) ? klass.primary_key.collect{|c|c.to_sym}.include?(col.name.to_sym) : col.name.to_sym == klass.primary_key.to_sym)
 
-        col_type = (col.type || col.sql_type).to_s
+        col_type = (col.sql_type || col.type).to_s
         if col_type == "decimal"
           col_type << "(#{col.precision}, #{col.scale})"
         elsif col_type != "spatial"
--- a/spec/annotate/annotate_models_spec.rb
+++ b/spec/annotate/annotate_models_spec.rb
@@ -28,7 +28,7 @@ describe AnnotateModels do
 
     stubs = default_options.dup
     stubs.merge!(options)
-    stubs.merge!(:name => name, :type => type)
+    stubs.merge!(:name => name, :sql_type => type, :type => type)
 
     double("Column", stubs)
   end


ctran:
A small change but with big impact.  I'm not sure how other would like this.

vassilevsky:
I think that people will like more correct annotations.

alexch:
I'm intrigued. How is a point specified in an AR schema or migration?

As a compromise, we could add the sql_type in parens if it's different from
the AR type.

ctran:
What I meant is I don't know much this will change over the existing annotations.  @alexch's suggestion make sense if you don't mind making that change.

vassilevsky:
I tested it on Rails 4.2.1 now.

```sh
rails new maps -d postgresql -T
cd maps
rails g model marker name location:point
```

```ruby
class CreateMarkers < ActiveRecord::Migration
  def change
    create_table :markers do |t|
      t.string :name
      t.point :location

      t.timestamps null: false
    end
  end
end
```

```sh
rake db:create db:migrate
pgcli maps_development 
```

```
maps_development> \d markers
+------------+-----------------------------+------------------------------------------------------+
| Column     | Type                        | Modifiers                                            |
|------------+-----------------------------+------------------------------------------------------|
| id         | integer                     | not null default nextval('markers_id_seq'::regclass) |
| name       | character varying           |                                                      |
| location   | point                       |                                                      |
| created_at | timestamp without time zone | not null                                             |
| updated_at | timestamp without time zone | not null                                             |
+------------+-----------------------------+------------------------------------------------------+
```

```ruby
# ...

ActiveRecord::Schema.define(version: 20150412093810) do

  # These are extensions that must be enabled in order to support this database
  enable_extension "plpgsql"

  create_table "markers", force: :cascade do |t|
    t.string   "name"
    t.point    "location"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
  end

end
```

I think that the proposed change will not cause much trouble in people's model annotations. First of all, not many people use those exotic types :) And if they do, and their `string`s change to `point`s, then, after a brief wtf moment, they should feel better informed and therefore happy. Right?

vassilevsky:
Thank you.

PR #240, opened 2015-03-20
Test on Ruby 2 versions

nicolasleger:


ctran:
Thanks for trying.  I've temporarily disabled the integration tests due to errors in Travis.  If you would like to fix those, that would be tremendously helpful as well.

Issue #236, opened 2015-03-12
Request: foreign key support


noinkling:
Now that Rails (>4.2) has support for database foreign keys, it would be nice to have an option for seeing these in the annotations.

noinkling:
`ActiveRecord::Base.connection.supports_foreign_keys?` can be used to check if the db supports them (I think only sqlite doesn't), then `ActiveRecord::Base.connection.foreign_keys("table_name")` can be used to get a list of them for any particular table.

franklsf95:
+1 would be really convenient!

PR #241, opened 2015-03-27
Adds support for foreign key annotations

subakva:
This adds support for foreign key annotations as requested in issue #236 .

I tried to follow the pattern used for index annotation, so it's enabled by default, and adds an additional section below the column definitions.

There's an example of the format in the spec: https://github.com/subakva/annotate_models/blob/features/foreign_key_annotations/spec/annotate/annotate_models_spec.rb#L160


ctran:
Thanks!!!

mhenrixon:
Can we get a release with this?

ctran:
I'll do a release this Thursday.

Issue #167, opened 2014-03-04
# -*- SkipSchemaAnnotations can not work .


songjiayang:
error message `Unable to annotate user.rb: stack level too deep` when i run migrate .
i add `# -*- SkipSchemaAnnotations` in user.rb, but it not work yet , but 
`skip_on_db_migrate=1 rake db:migrate` work well . thanks

gbenedict:
I'm having the same issue with `# -*- SkipSchemaAnnotations` – it doesn't skip the annotation.

ctran:
Please provide more details: your version of ruby and annotate, Gem file, any other useful information.
Also the symptoms of why you think it doesn't work.

gbenedict:
ruby: 2.1.2p95
gem: 2.6.5

I have added `# -*- SkipSchemaAnnotations` to several models, but the annotations are still added to those models during migrations or running annotate directly. The models I am trying to skip are in a subdirectory for organization. They also use STI and inheritance (`WorkflowActionCallExternalApi < WorkflowAction`).

PR #242, opened 2015-03-29
Update annotate_models.rb

mafernando:
Potentially addresses issue# https://github.com/ctran/annotate_models/issues/167
Annotate_model_file is called automatically on all files in models directory and does not check if model contains SkipSchemaAnnotations tag.


mafernando:
Nevermind, need to understand how do_annotate works better. Sorry for the noise.

PR #243, opened 2015-03-30
Check if model has skip tag in annotate_model_file

mafernando:
Potentially addresses issue# #167
Annotate_model_file is called automatically on all files in models directory and does not check if model contains SkipSchemaAnnotations tag.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -477,6 +477,7 @@ module AnnotateModels
 
     def annotate_model_file(annotated, file, header, options)
       begin
+        return false if (/# -\*- SkipSchemaAnnotations.*/ =~ (File.exist?(file) ? File.read(file) : '') )
         klass = get_model_class(file)
         if klass && klass < ActiveRecord::Base && !klass.abstract_class? && klass.table_exists?
           if annotate(klass, file, header, options)


mafernando:
In my scenario, I have a model file (more of an api wrapper) which doesn't have an actual DB table associated with it. I tried adding the `# -*- SkipSchemaAnnotations` tag to it, but annotate sill generates the following error message with `rake db:migrate`:

```
Unable to annotate app/models/client.rb: cannot load a model from app/models/client.rb
```

I attached a debugger on [`AnnotateModels.do_annotations(options)`](https://github.com/mafernando/annotate_models/blob/develop/lib/tasks/annotate_models.rake#L37) and traced through to [`annotate_models.rb:do_annotations`](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L455) > [`annotate_models.rb:annoatate_model_file`](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L469) and noticed that annotate checks if [`klass.table_exists?`](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L481) when it is iterating through the list of model files to [`annotate`](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L482) and throws an error if it does not exist. This [error](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L487) is thrown before the actual [`annotate`](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L342) method is called, so the [skip tag check](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L266) that would be invoked in [`annotate_one_file`](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L263) does not get called.  

In summary, I'm suggesting that a skip tag check be added to [`annotate_model_file`](https://github.com/mafernando/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L478) so that annotate checks to see if the model file has a skip tag prior to checking if its table exists which will cause an error to be thrown as it currently is.

Here is the `auto_annotate_models.rake` file that I am using with my app.

```ruby
# NOTE: only doing this in development as some production environments (Heroku)
# NOTE: are sensitive to local FS writes, and besides -- it's just not proper
# NOTE: to have a dev-mode tool do its thing in production.
if Rails.env.development?
  task :set_annotation_options do
    # same name.
    Annotate.set_defaults(
      'position_in_routes'   => 'before',
      'position_in_class'    => 'before',
      'position_in_test'     => 'before',
      'position_in_fixture'  => 'before',
      'position_in_factory'  => 'before',
      'show_indexes'         => 'true',
      'simple_indexes'       => 'false',
      'model_dir'            => 'app/models',
      'include_version'      => 'false',
      'require'              => '',
      'exclude_tests'        => 'false',
      'exclude_fixtures'     => 'false',
      'exclude_factories'    => 'false',
      'ignore_model_sub_dir' => 'true',
      'skip_on_db_migrate'   => 'false',
      'format_bare'          => 'true',
      'format_rdoc'          => 'false',
      'format_markdown'      => 'false',
      'sort'                 => 'false',
      'force'                => 'false',
      'trace'                => 'false'
    )
  end

  Annotate.load_tasks unless
end
```

Issue #244, opened 2015-04-01
exclude_serializers option is not honored from the rake task


shuriu:
As the title says, if I add an option to exclude serializers: `"exclude_serializers" => "true"'`to the auto generated rake task, it updates the serializers as well.

Using:
Ruby: `2.2.1p85`
Rails: `4.2.1`
Annotate: `2.6.8`

PR #246, opened 2015-04-10
fix issue where serializer-related flags weren't being honored

bruschill:
As mentioned in #244, serializer-specific flags were being ignored regardless of if they were specified in the generated `auto_annotate_models.rake` file or if they were passed via CLI. This fixes that.


ctran:
Please try to avoid reformatting code that is not directly needed by the PR.  It makes it harder for me to see the actual changes :-)

bruschill:
My apologies! This was the first pull request I've ever submitted, so I'll keep it in mind for the future.

PR #250, opened 2015-05-25
ignore default value for jsonb

LcpMarvel:


ctran:
Could you remind me why you wanted this change?

LcpMarvel:
The default value of jsonb, which is longer than other types, makes the schema Information ugly.

PR #254, opened 2015-06-03
Invoke get options task before annotating migrations

### LINKED ISSUE #252: How do I turn off annotations during migrate?
`'skip_on_db_migrate' => 'false'`  seemingly does not help and every time I migrate and I have to run `bundle exec annotate` to refresh the indexes. Otherwise my foreign keys and such is lost.

arathunku:
As in title.

[ fixes #252 ]


ctran:
Thanks!

PR #257, opened 2015-06-15
Quote each value in default arrays

jmondo:
Previously, quoting an array fell back to `inspect`, which is ok for
strings, but not good for BigDecimals, for instance.


jmondo:
in our project we have an order model with "individual portions" column that is an array of decimals in postgres. the default is an array with the BigDecimal 1.0 in it.

before: 
```
#  individual_portions                        :decimal(, )      default([#<BigDecimal:7fd89b22c1a0,'0.1E1',9(18)>]), is an Array
```
after: 
```
#  individual_portions                        :decimal(, )      default(["1.0"]), is an Array
```

Issue #266, opened 2015-07-14
Annotate does not accept wrapper default parameters.


Crote:
Annotate::set_defaults(options = {}) filters options by POSITION_OPTIONS, FLAG_OPTIONS, PATH_OPTIONS, OTHER_OPTIONS, none of which include `wrapper`, `wrapper_open` and `wrapper_close`. This makes it impossible to use those options from the generated rake task. Please fix this.

PR #283, opened 2015-10-26
Accept wrapper default params

kamilbielawski:
This PR allows setting default `wrapper_start` and `wrapper_end` options for rake task.
I didn't include `wrapper` option here. While it's useful as a shorthand when setting options directly from the command line, I think it might be unnecessarily confusing here. If you don't agree, I can include it in the rake task as well.

PR fixes issue #266

### DIFF
--- a/lib/annotate.rb
+++ b/lib/annotate.rb
@@ -29,7 +29,7 @@ module Annotate
     :timestamp, :exclude_serializers, :classified_sort, :show_foreign_keys,
   ]
   OTHER_OPTIONS=[
-    :ignore_columns, :skip_on_db_migrate
+    :ignore_columns, :skip_on_db_migrate, :wrapper_open, :wrapper_close
   ]
   PATH_OPTIONS=[
     :require, :model_dir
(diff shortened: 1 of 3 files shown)


kamilbielawski:
Hmm, it looks like Semaphore isn't configured properly, doesn't it?

lib/annotate.rb
```diff
@@ -78,6 +78,12 @@ def self.setup_options(options = {})
     return options
   end
 
+  def self.reset_options
+    [POSITION_OPTIONS, FLAG_OPTIONS, PATH_OPTIONS, OTHER_OPTIONS].flatten.each do |key|
+      ENV[key.to_s] = nil
```
ctran: What is the reason to set these to nil?

ctran:
Please ignore the failure from semaphoreci, I didn't have the correct configuration.

lib/annotate.rb
kamilbielawski: It's needed for the tests - ENV wasn't cleared between tests, so after wrapper attributes were set in one of the tests, the others also had it set, which resulted in incorrect annotations being generated.

ctran:
Thanks for the PR!

Issue #286, opened 2015-11-11
annotate additional subdirectories 


djsegal:
a scaffold generates files in the following rspec subdirectories:
  + `spec/models`
  + `spec/controllers`
  + `spec/requests`
  + `spec/routing`

currently, only `spec/models` is annotated. is there a way to annotate the other three?

ctran:
A PR is welcome!  What version of rails does this?

djsegal:
I'm testing it now. I'm working with a Rails 5 API

djsegal:
I just finished the PR and tested it with test_unit and rspec: https://github.com/ctran/annotate_models/pull/287

djsegal:
@ctran is there anything more I can do to help?

ctran:
Sorry, I will take a look tomorrow.

ctran:
This is now merged into "develop".  It will be part of the next release.

PR #287, opened 2015-11-11
Annotate non-model test files: #286

djsegal:
----------

Related to #286:

----------

a scaffold generates files in the following rspec subdirectories:

+ `spec/models`
+ `spec/controllers`
+ `spec/requests`
+ `spec/routing`

currently, only `spec/models` is annotated. is there a way to annotate the other three?


ctran:
Thanks!!!

PR #289, opened 2015-11-17
rubocop friendly template

roflomglol:
Hey,

I've started to use your gem and my rubocop exploded. :) Probably I'm not the only one. I know that single-qouted or double-quoted string are more likely a matter of taste, but at least preserve consistency. And I've removed curly braces around hash parameter, 1.9.3 supports that.

Thanks,
Sergey


ctran:
Could you resubmit against the latest of develop branch?  Thanks

lib/generators/annotate/templates/auto_annotate_models.rake
```diff
@@ -5,34 +5,34 @@ if Rails.env.development?
...
     # You can override any of these by setting an environment variable of the
     # same name.
-    Annotate.set_defaults({
-      'position_in_routes'      => "before",
-      'position_in_class'       => "before",
-      'position_in_test'        => "before",
-      'position_in_fixture'     => "before",
-      'position_in_factory'     => "before",
-      'position_in_serializer'  => "before",
-      'show_foreign_keys'       => "true",
```
Crote: TRW

Issue #291, opened 2015-11-22
Annotations for models in subdirectories of engines/subtrees/etc


djsegal:
I believe this is similar to the following issues: #190, #131, #88

To get it to work on my project, I have the following 2 lines in `auto_annotate_models.rake`:

```
 'model_dir'            => ["app/models", "lib/subtree/app/models"],
 'root_dir'             => ["/", "lib/subtree"],
```

djsegal:
This has been merged.

Issue #190, opened 2014-05-31
Annotation of models in Rails Engine


bobbdelsol:
Hey, as best as i can tell, you don't recognize models inside a rails engine.

So the module Enginename    Wrapper is throwing you off.

Also, they're pushed down on level in models  models/engingname/....

OneDivZero:
Yes that would be a really nice feature!

cheef:
+1

ctran:
It should now work.

Issue #131, opened 2013-06-02
annotate my rails engine models


paulopatto:
What if I need to write down the models of a rails engine? I use to annotate the application that makes use of a rails engine.

Issue #88, opened 2012-07-29
doesn't work in engines ?


alotela:
Hi,

is it me, or annotate doesn't work with engines models ?

thanks.

PR #292, opened 2015-11-22
Add root_dir capability for multiple roots

djsegal:
---------------

Annotations for models in subdirectories of engines/subtrees/etc #291

---------------

I believe this is similar to the following issues: #190, #131, #88

To get it to work on my project, I have the following 2 lines in auto_annotate_models.rake:

```
 'model_dir'            => ["app/models", "lib/subtree/app/models"],
 'root_dir'             => ["/", "lib/subtree"],
```

### DIFF
--- a/lib/annotate.rb
+++ b/lib/annotate.rb
@@ -33,7 +33,7 @@ module Annotate
     :ignore_columns, :skip_on_db_migrate, :wrapper_open, :wrapper_close, :wrapper
   ]
   PATH_OPTIONS=[
-    :require, :model_dir
+    :require, :model_dir, :root_dir
   ]
 
 
@@ -74,6 +74,10 @@ module Annotate
 
     if(options[:model_dir].empty?)
       options[:model_dir] = ['app/models']
+    end
+
+    if(options[:root_dir].empty?)
+      options[:root_dir] = ['']
     end
 
     options[:wrapper_open] ||= options[:wrapper]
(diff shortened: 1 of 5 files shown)


lib/annotate.rb
```diff
@@ -76,6 +76,10 @@ def self.setup_options(options = {})
       options[:model_dir] = ['app/models']
     end
 
+    if(options[:root_dir].empty?)
+      options[:root_dir] = ['/']
```
ctran: Is this supposed to be Rails.root? A slash doesn't seem to a correct default value from the caller/user perspective.

djsegal: + Thats true, we could just make the default `''`
+ The place where root_dir is used in will be unchanged:
```
File.join(root_directory, UNIT_TEST_DIR,  "%MODEL_NAME%_test.rb")
```

ctran: Could you make the change so I can accept the PR?

djsegal:
@ctran I just finished the update. (i'll clean up the other PRs after the merge)

--------

One note, I did some minor refactoring:
+ I added `MATCHED_TYPES = %w(test fixture factory serializer scaffold)` to top of `lib/annotate/annotate_models.rb`
+ i default `get_patterns` to use `MATCHED_TYPES`, i.e. `get_patterns(pattern_types=MATCHED_TYPES)`

ctran:
Awesome. Thanks.

Issue #293, opened 2015-11-22
Annotations on the actual controller files


djsegal:
This was the last generated file that wasn't annotated, so I thought we should add it?

This would be like `app/controllers/users_controller.rb`

PR #294, opened 2015-11-22
Annotate controllers for models

djsegal:
--------

Annotations on the actual controller files #293 

---------

This was the last generated file that wasn't annotated, so I thought we should add it?

This would be like `app/controllers/users_controller.rb`

------

*Note: this is built off the branch created for issue #291*


djsegal:
this needs a refactor after #292 goes live

ctran:
I'm closing this one.

PR #298, opened 2015-11-30
Fix :ignore_columns option in "auto_annotate_models" task

wouterw:
When using the `auto_annotate_models` task, the `ignore_columns` option was always set to `""`, which caused `AnnotateModels.do_annotations` to ignore all columns.


ctran:
It would seem this would benefit from some tests :-)

PR #299, opened 2015-12-02
Annotate controllers and helpers

### LINKED ISSUE #293: Annotations on the actual controller files

djsegal:
##### Resolves #293:

I thought it might be useful to add annotations to helpers and controllers.

The next question is do we allow everything that can be generated to be annotated?

-------

This includes:

+ javascript assets
+ stylesheet assets
+ migrations themselves
+ [js/html/json] . erb/jbuilder files for
  * index
  * show
  * edit
  * _form

### DIFF
--- a/lib/annotate.rb
+++ b/lib/annotate.rb
@@ -27,7 +27,7 @@ module Annotate
     :exclude_fixtures, :exclude_factories, :ignore_model_sub_dir,
     :format_bare, :format_rdoc, :format_markdown, :sort, :force, :trace,
     :timestamp, :exclude_serializers, :classified_sort, :show_foreign_keys,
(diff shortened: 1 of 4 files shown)


ctran:
I don't think it makes sense to annotate everything, especially when they are not "rb" files.  I'm already not sure about controllers and helpers.  Do you find them helpful?  What do others think?

djsegal:
@ctran:
+ one reason I use annotate is because its a subtle reminder to update every file related to a model. 
+ i do agree that some files may be overkill, i just wanted to see if annotate should be able to touch everything made from a `rails g blank ....`

PR #302, opened 2015-12-10
Ruby 2.3 magic comments support

b264:
My attempt to support Ruby 2.3 magic comments

### DIFF
--- a/spec/annotate/annotate_models_spec.rb
+++ b/spec/annotate/annotate_models_spec.rb
@@ -502,14 +502,19 @@ end
       Annotate::PATH_OPTIONS.each { |key| ENV[key.to_s] = '' }
     end
 
-    def encoding_comments_list_each
+    def magic_comments_list_each
       [
         '# encoding: UTF-8',
         '# coding: UTF-8',
         '# -*- coding: UTF-8 -*-',
         '#encoding: utf-8',
-        '# -*- encoding : utf-8 -*-'
-      ].each{|encoding_comment| yield encoding_comment }
+        '# encoding: utf-8',
+        '# -*- encoding : utf-8 -*-',
+        "# encoding: utf-8\n# frozen_string_literal: true",
+        "# frozen_string_literal: true\n# encoding: utf-8",
+        '# frozen_string_literal: true',
+        '#frozen_string_literal: true',
+      ].each{|magic_comment| yield magic_comment }
     end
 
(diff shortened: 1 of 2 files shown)


ctran:
I haven't heard of this.  Where can I find more info on this?

b264:
nithinbekal.com/posts/ruby-2-3-features/

b264:
Also I tried the current version with Ruby `2.3.0preview1`

spec/annotate/annotate_models_spec.rb
```diff
@@ -502,14 +502,19 @@ def annotate_one_file options = {}
...
         '# coding: UTF-8',
         '# -*- coding: UTF-8 -*-',
         '#encoding: utf-8',
-        '# -*- encoding : utf-8 -*-'
-      ].each{|encoding_comment| yield encoding_comment }
+        '# encoding: utf-8',
+        '# -*- encoding : utf-8 -*-',
+        "# encoding: utf-8\n# frozen_string_literal: true",
+        "# frozen_string_literal: true\n# encoding: utf-8",
+        '# frozen_string_literal: true',
```
ctran: Am I reading this correctly?  These seem to be repeated.

ctran:
This seems like an optional feature since some people may not want to turn it on.  I think we have an option to append additional text so that may be a better option.  Can you enlighten me?

b264:
If you have the `# encoding: utf-8` magic comment (aka directive pragma) at the top of the file, annotate already leaves this there. :+1:  Ruby 2.3.0 will have an additional magic comment now which I for one will be using immediately.  This PR will cause annotate to leave that magic comment alone at the top of the file, as it leaves the encoding one there also.  I thought I had read somewhere that it had to be on the first 2 lines, but after additional searching and an experiment, it seems at least in the preview the magic comment can be anywhere before the first executable line and still affect the compile mode.  However, moving it from the top of the file in my opinion is still superbly messy and makes it harder to know you're working in frozen string land.  So the entire point is that it leaves it at the top of the file, like the other one, instead of burying it under the schema.  This does not turn the feature on if they hadn't already turned it on; it preserves their action to activate it.

spec/annotate/annotate_models_spec.rb
b264: I don't think any are duplicated

ctran:
Thanks

PR #310, opened 2016-01-08
fix code markup in docs

hedgesky:
RDoc supports +code+ syntax only for words, not phrases. That's why several commands in Readme are not rendered correctly. This PR fixes the issue by using <code> tag there.


ctran:
Thanks.

PR #312, opened 2016-01-09
Add ignore_routes option

mintuhouse:
I use to skip routes which generate an annotation which includes local path e.g.,
````
#     /mobile       #<Rack::Builder:0x0000000a2ff800 @run=Mobile::Dispatch, @map=nil, @use=[#<Proc:0x0000000a30e2d8@/home/hasan/.rvm/gems/ruby-2.1.7@gogo/gems/rack-1.5.5/lib/rack/builder.rb:86>]>
#     /resque       #<Resque::Server app_file="/home/hasan/.rvm/gems/ruby-2.1.7@gogo/gems/resque-1.25.2/lib/resque/server.rb">
````
which causes a file change after every annotate run - leading to bad churn on codeclimate

add `ignore_routes: "#<"` to config options to skip such routes

### DIFF
--- a/bin/annotate
+++ b/bin/annotate
@@ -178,6 +178,10 @@ OptionParser.new do |opts|
     ENV['ignore_columns'] = regex
   end
 
+  opts.on('-I', '--ignore-routes REGEX', "don't annotate routes that match a given REGEX (i.e., `annotate -I '(mobile|resque|pghero)'`" ) do |regex|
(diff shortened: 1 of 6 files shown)


ctran:
Thanks

mintuhouse:
https://github.com/ctran/annotate_models/pull/315
resubmitted against develop

PR #319, opened 2016-01-15
Added Active admin models.

ryanfox1985:
Hi,

I added the active admin models folder.

Regards!


sethdeckard:
How do you disable this?

PR #328, opened 2016-01-20
Fix some rubocop offences

b264:


ctran:
Could you rebase these into a single commit?

ctran:
No longer needed because of #332

PR #332, opened 2016-01-25
Reduce rubocop offenses and don't include activeadmin models by default

### LINKED ISSUE #330: No way to disable annotating Active Admin models
I want my models annotated by my active admin models left alone. Unless I'm just missing it, there doesn't seem to be a way to disable this.

ryanfox1985:
Reduced the Cyclomatic complexity.
#330 By default didn't annotate activeadmin models.
This close #330


ryanfox1985:
One moment I think not work.

Issue #329, opened 2016-01-22
Wrapper open persists after deleting annotations


sethdeckard:
This means if you run the command to add and delete annotations multiple times you end up with something like this in your files:

    # rubocop:disable Metrics/LineLength

    # rubocop:disable Metrics/LineLength

    # rubocop:disable Metrics/LineLength

    # rubocop:disable Metrics/LineLength

    # rubocop:disable Metrics/LineLength

    # rubocop:disable Metrics/LineLength

PR #339, opened 2016-01-31
Don't leave wrapper when removing annotations

kamilbielawski:
This fixes issue #329 .
When removing annotations it will also remove the opening wrapper.
There are two ways it could be implemented:
* remove line before the annotation if it matches the `wrapper_open`
* remove the line before the annotation if it's a comment

I went with the first option, because I wanted to avoid the risk of removing a comment that isn't a wrapper opening. The drawback is that it will only work if the `wrapper_option` is set to the same value as it was when the annotation was created.
Let me know if you think the alternative approach would be better.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -403,10 +403,11 @@ module AnnotateModels
       end
     end
 
-    def remove_annotation_of_file(file_name)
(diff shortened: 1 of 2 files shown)


ctran:
Thanks for providing the tests.

PR #347, opened 2016-02-18
Fix fatal error when using `annotate` with ActiveRecord 4.1

brentwalther:
The fatal error was introduced with commit fc31d85. The method
`supports_foreign_keys?` is only available on ActiveRecord versions
`>= 4.2`.

Fixes issue 316.


ctran:
Thanks.

PR #351, opened 2016-02-26
Allow classified sorting without an ID column

### LINKED ISSUE #348: classified sort has problems if no id is present
Hi,

I want to switch from ```annotate``` to ```annotate --classified-sort``` and I encounter the following problem:

```
Unable to annotate app/models/schema_migration.rb: undefined method `type' for nil:NilClass
```

It also happens to other models without an ID too and the file gets not annotated at all. Without ```--classified-sort``` it works as expected, but the columns are obviously not sorted.

kfarst:
Resolves #348


ctran:
Thanks!

PR #353, opened 2016-03-01
Add unit tests for classified sort option

kfarst:
@ctran Added a unit test to cover the fix for https://github.com/ctran/annotate_models/issues/348


ctran:
Love tests.  Thanks!

PR #358, opened 2016-03-22
Annotate on_delete/on_update foreign key constraints

sashabelozerov:
I find it very useful to have `on_delete` and `on_update` FK constraint values in annotations.

e.g.

`fk_rails_02e851e3b7  (foreign_thing_id => foreign_things.id) ON DELETE => nullify ON UPDATE => cascade`

in case of `-f makrdown` option:

`fk_rails_02e851e3b7 ` (_ON DELETE => nullify ON UPDATE => cascade_)


ctran:
Thanks!

PR #359, opened 2016-03-27
Allow rake 11 in annotate.gemspec

pwnall:


pwnall:
I manually tested both this PR and the PR against `develop` using a Rails 5 app. Things appear to be working just fine with Rake 11.

ctran:
Thanks.

pwnall:
Thank you very much for the quick turnaround! :+1:

Issue #357, opened 2016-03-17
annotate --routes changes model files


aminariana:
When I do a simple:
`annotate --routes`

I expect only the `routes.rb` file to change, not the model annotations. But they do.

My usage hasn't changed from previous versions of annotate. But going from `v2.6.10` to `v2.7`, when I run the command, it overrides my existing model annotations with fresh ones, without the desired options such as annotating indexes that I otherwise specify in annotating models. Breaking change I think?

EDIT:

Downgrading to v2.6.10 fixes the problem.

kamilbielawski:
I believe the behaviour you described is like that by design.
I've just made a small PR with additional option (`--ignore-models`) that will let you achieve desired behaviour. However, I'm not entirely sure if that's the right approach. It would be great to hear @ctran's opinion

aminariana:
If by design, why inconsistent between two versions?

If I annotate routes, it violates my expectations to see model files modified.

Hitabis:
Why not annotating only the routes when calling `annotate --routes`?
It would only be necessary to check the presence of the `routes` argument and the user would get the expected result.
This is the function causing the unexpected behavior: https://github.com/ctran/annotate_models/blob/6775e4a16d853c3484f7753198220cbeb99bba8d/lib/annotate.rb#L116

ctran:
Ah, I misread the existing behavior.

javierguajardo:
@aminariana I've tested your issue using:
* Rails 5.2.0
* Ruby 2.5.1
* Annotate 2.7.3

I ran `annotate --routes` and only the `config/routes.rb` file was changed. I believe this problem has been patched with 2.7.3.

Regards.

PR #361, opened 2016-03-30
Added --ignore-models option for skipping models annotation

kamilbielawski:
Solves #357

### DIFF
--- a/bin/annotate
+++ b/bin/annotate
@@ -128,6 +128,11 @@ OptionParser.new do |opts|
     ENV['root_dir'] = dir
   end
 
+  opts.on('--ignore-models',
+          "Don't annotate models") do |dir|
+    ENV['ignore_models'] = 'yes'
+  end
+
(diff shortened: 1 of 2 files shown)


ctran:
Thanks for the PR but I was thinking of moving away from using the "exclude" options and use a "types" flag to allow individual types to be annotated.  It would have a default list of "types" to preserve the current behavior.  Want to implement that?

kamilbielawski:
Yeah, I wasn't really convinced that adding this `--ignore-models` switch is a good idea but it was the simplest way to achieve functionality requested in the issue #357 .
I can implement `--types` flag, although it might take a while as I don't know when I find some time to do it.

Hitabis:
Is this expected to be merged into the master branch? I'm also having the same problem described in #357 and would be happy to see this released soon.

ctran:
The master branch merge will only happen when I do a release.  Can you point to the dev branch?

PR #365, opened 2016-04-21
add option to exclude STI subclasses from annotation

ethnext:
I've added an option named `exclude_sti_subclasses` and code which checks if a model has a superclass under AR::B with the same table in order to exclude STI subclasses from annotation. 

this is a need I have on a project with a large number of subclasses of certain models, where annotating every one generates a lot of noise on any schema change, so I prefer to just annotate the parent class.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -105,7 +105,7 @@ Metrics/MethodLength:
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 517
+  Max: 522
 
 # Offense count: 7
 Metrics/PerceivedComplexity:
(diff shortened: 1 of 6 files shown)


ecnalyr:
Is there a way to exclude a single file from the exclude_sti_subclasses option?

ethnext:
no, not as this is currently implemented

ecnalyr:
This doesn't look like it breaks anything: https://github.com/ctran/annotate_models/pull/366

ctran:
Can you fix the broken CI tests?

ethnext:
not sure about the semaphore failure now. it doesn't look related to changes I made.

ctran:
Thanks!

PR #370, opened 2016-04-30
Add --types option

kamilbielawski:


ctran:
Thanks for doing this.  I'll spend some time looking at the changes.

lib/generators/annotate/templates/auto_annotate_models.rake
```diff
@@ -13,6 +13,16 @@ if Rails.env.development?
       'position_in_fixture'     => 'before',
       'position_in_factory'     => 'before',
       'position_in_serializer'  => 'before',
+      'include_model'           => true,
```
ctran: Since we're introducing the "types" option, why don't we use that here as well instead of having these?  Something similar to what you did above:

'types' => 'model,test,fixture,factory,serializer,scaffold,controller,helper'
or
'types' => %w(model test fixture factory serializer scaffold controller helper)

I need to figure out what the previous defaults since the new defaults seem to surprise quite a number of people (I can do that after this PR if necessary).

I think you're on the right track but I'll need to check this out on my machine to play with it a bit.

kamilbielawski: Heh, what you're suggesting here is exactly how I first implemented it. I changed it later to be more consistent with existing options (like `exclude_*`). I'll go back to the previous approach.

kamilbielawski:
@ctran Did you have a chance to look at my changes? Please let me know if there are still some improvements needed

ctran:
Yes, there are changes I want to make.  You can take a look at the
types-flag branch.

kamilbielawski:
OK, so basically a cleanup of the obsolete options, right?

ctran:
Pretty much.

PR #372, opened 2016-05-10
Add new ruby versions

ctran:


b264:
Do you think 2.3.1 is worth including?

Issue #377, opened 2016-05-18
PG Expression Indexes Produce Undefined Method `join` for String


bousquet:
I have a case-insensitive index on `User#username` which produces the following index definition:

```ruby
#<struct ActiveRecord::ConnectionAdapters::IndexDefinition table="users", name="index_users_on_lowercase_username", unique=true, columns="lower((username)::text)", lengths=[], orders=nil, where=nil, type=nil, using=:btree, comment=nil>
```

Because the `#columns` attribute of that struct is a String, we get the following errors when annotating our model:

```plain_text
Unable to annotate app/models/user.rb: undefined method `join' for "lower((username)::text)":String
Did you mean?  JSON
Unable to annotate app/models/user.rb: no implicit conversion of nil into Array
Model files unchanged.
```

PR #378, opened 2016-05-18
Coerce string `index.columns` to Array before join

bousquet:
Fixes Issue #377 

Tested manually on Rails 5.0.0.rc1.  String and array indexes work with this patch.  I didn't see any existing place in the tests to add this.


ctran:
Thanks!

PR #379, opened 2016-05-26
Fix spelling to commas

gregsherrid:
Command line options documentation previously asked for values to be separated by "comas" not "commas."


ctran:
Thanks!

PR #385, opened 2016-06-30
Add --additional-subdir option

nicklandgrebe:
This adds a path option called `--additional-subdir`, which according to the description, allows one to "Additionally annotate files in subdir of each scaffold/serializer/controller/helper dir, separate multiple dirs with comas"

The reason I made this feature is because in my app, I have a bunch of models, but each of these models can have multiple serializers, for example: `serializers/products_serializer.rb` and `serializers/api/v1/products_serializer.rb` both exist for the `Product` model. I do not want to namespace the `Product` model, because we are simply providing two different interfaces for the same model.

To accomplish this, we make the call `annotate --additional-subdir api/v1` and it works great!


ctran:
What's going on? There are 74 commits on this PR.

nicklandgrebe:
I did a rebase and it must have messed up. Let me try again and I'll get back to you.

ctran:
Thanks!

nicklandgrebe:
#386 sorry about that!

PR #386, opened 2016-07-02
Add additional-subdir option

nicklandgrebe:
This adds a path option called --additional-subdir, which according to the description, allows one to "Additionally annotate files in subdir of each scaffold/serializer/controller/helper dir, separate multiple dirs with comas"

The reason I made this feature is because in my app, I have a bunch of models, but each of these models can have multiple serializers, for example: serializers/products_serializer.rb and serializers/api/v1/products_serializer.rb both exist for the Product model. I do not want to namespace the Product model, because we are simply providing two different interfaces for the same model.

To accomplish this, we make the call annotate --additional-subdir api/v1 and it works great!

### DIFF
--- a/bin/annotate
+++ b/bin/annotate
@@ -189,7 +189,7 @@ OptionParser.new do |opts|
     ENV['ignore_unknown_models'] = 'true'
   end
 
-  opts.on('--additional-subdir dir', "Additionally annotate files in subdir of each scaffold/serializer/controller/helper dir, separate multiple dirs with comas" ) do |dir|
+  opts.on('--additional-subdir dir', "Additionally annotate files in subdir of each scaffold/serializer/controller/helper dir, separate multiple dirs with comas") do |dir|
     ENV['additional_subdir'] = "#{dir}"
   end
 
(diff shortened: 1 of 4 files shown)


ctran:
Have you looked at the root_dir option (#292)?  It seems to be very similar to what you're proposing here.

nicklandgrebe:
It's really close, but even if I were to switch my repo to its intended use case by:

Taking all my subdirs and moving them to another root folder using a Rails Engine:

```
/app/controllers/api/v1/*.rb
->
/api/v1/app/controllers/api/v1/*.rb`
```

You'll note the engine still requires that I put all my controllers in module `API::V1` that corresponds to a folder subdirectory `controllers/api/v1/`, so I still need this feature even though `root_dir` outside a Rails application would work here.

pelletencate:
I need this one. Is this going to be merged soon?

ctran:
Please rebase if you still want me to merge this.

drwl:
@nicklandgrebe gonna close this as this PR is quite old. Feel free to reopen if you would like to update it and get it merged in.

Issue #320, opened 2016-01-15
Default value for JSON datatype breaks annotation


xn:
I was using version 2.6.something. Upgraded to 2.7.0 and still have a problem.
```
# == Schema Information
#
# Table name: users
#
#  id                     :integer          not null, primary key
#  profile              :json             default({"first_name"=>nil,
 "last_name"=>nil,
 "phone"=>nil,
 "terms_of_service"=>nil,
 "street"=>nil,
 "suite"=>nil,
 "city"=>nil,
 "state"=>nil,
 "zip"=>nil}
)
```

ctran:
You'll have to do some debugging to help me out.  This seems to be working for others #305 in 2.7.0.  This was handled by #250.

xn:
Just got this, awesome. thanks.

PR #389, opened 2016-07-18
Provide :hide_default_column_types option to make hiding of default configurable

jastkand:
First of all, thanks for the great gem!

`2.7.0` release made hidden the default value of `hstore`, `json` and `jsonb` columns. It's reasonable in such cases like this: https://github.com/ctran/annotate_models/issues/320. But that's not good when the default value is more simple, like `default({})` or `default([])`. I think it would be great to have hiding of `default` configurable. This pull-request introduces this feature.


ctran:
Great!  One question though,  I didn't see the existing defaults of "json,jsonb,hstore" being used if the option is not specified.  Did I miss it somewhere?

jastkand:
The goal is to be able to disable hiding of default. Using default constant when the option value is blank does not allow to switch hiding off. So in current implementation user should explicitly specify "json,jsonb,hstore" which is not good.

I'm thinking of updating the code to use default constant when the option is `nil` and not to use it when the option is empty string. @ctran What do you think?

jastkand:
@ctran Would you mind taking a look at the PR?

ctran:
Thanks!!!

nneal:
I'm trying to use this in our environment, but doesn't seem to be working.

I have ```hide_default_column_types' => 'skip',  ``` as a setting in the annotate config file, i'm invoking annotate with `b exec rake annotate_models` and still not showing defaults for us :\ Not sure where I'm going wrong

jastkand:
@nneal do you use version of gem from master? The feature is not yes released in rubygems version of gem.

nneal:
I am using `gem 'annotate', git: 'https://github.com/ctran/annotate_models.git', branch: 'develop'` because it appears to be the most up to date branch on github, but i also tried `branch: 'master'`. Neither worked for these settings :\ Thank you for the help

nneal:
@jastkand I left more information in here: https://github.com/ctran/annotate_models/issues/441

Issue #334, opened 2016-01-26
2.7.0: controller tests are annotaded despite 'exclude_tests' => "true"


bughit:
other tests are not annotated

bughit:
caused by exclude_scaffolds

bughit:
one should not have to exclude a category twice (exclude_tests, exclude_scaffolds), exclude_tests should be sufficient for excluding tests

ctran:
This is still an issue for 2.7.2?

PR #395, opened 2016-08-30
Added option exlude_controller_tests

akito19:
Hi!
I use this gem whenever I start rails projects. However I was in trouble over to exclude annotating `/spec/controller/*` files until I've understood `exclude_scaffolds` option.

And furthermore, because I found an issue https://github.com/ctran/annotate_models/issues/334, I separated controller test/spec files from `AnnotateModels#scaffold_files` as `AnnotateModels#controller_test_files`.


ctran:
Please make sure all the tests are passing.

Issue #397, opened 2016-08-30
Should Ruby of old versions be dropped?


akito19:
I suggest that this gem may drop old ruby versions.
[Ruby 1.9.x and 2.0.0 have not already been supported](https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/).  Furthermore because `ruby_dep` gem, which requires Ruby >= 2.2.5, cannot been installed, most of tests fail.

PR #398, opened 2016-08-30
dropped ruby version < 2.2.5

akito19:
https://github.com/ctran/annotate_models/issues/397

I'm sorry for pull requests in succession. I made a commit that I dropped old versions of Ruby.
This PR can make some of fail tests of other PRs passed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 sudo: false
 language: ruby
 rvm:
-  - 1.9.3
-  - 2.0
   - 2.1
   - 2.2
-  - 2.2.3
   - 2.2.4
+  - 2.2.5
   - 2.3.0
+  - 2.3.1
   - ruby-head
 matrix:
   allow_failures:
@@ -16,6 +15,7 @@ matrix:
     - rvm: 2.0
     - rvm: 2.1
     - rvm: 2.2
+    - rvm: 2.3
 before_install:
   - rvm @global do gem install bundler
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ matrix:
     - rvm: 2.0
     - rvm: 2.1
     - rvm: 2.2
+    - rvm: 2.2.4
     - rvm: 2.3
 before_install:
   - rvm @global do gem install bundler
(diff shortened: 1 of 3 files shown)


akito19:
Can I SemaphoreCI rebuild with no cache?

.travis.yml
```diff
@@ -1,21 +1,16 @@
...
 matrix:
   allow_failures:
     - rvm: ruby-head
-    - rvm: 1.9.3
-    - rvm: 2.0
-    - rvm: 2.1
-    - rvm: 2.2
+    - rvm: 2.2.5
+    - rvm: 2.3.0
+    - rvm: 2.3.1
```
b264: I think only `ruby-head` and older, unsupported versions should be in the allow_failures section

b264: Having 1.9.3 in the allow-failures section can be useful if someone wants to know what parts will fail on 1.9.3

annotate.gemspec
```diff
@@ -7,7 +7,7 @@ Gem::Specification.new do |s|
   s.name = "annotate"
   s.version = Annotate.version
 
-  s.required_ruby_version = '>= 1.9.3'
+  s.required_ruby_version = '>= 2.2.5'
```
b264: Jumping all the way to 2.2.5 seems a little overkill.  The 2.1 series is still receiving security patches, so in my opinion it should probably stay supported unless there is an overwhelming benefit of dropping it.

akito19: If Ruby 2.1.x exist, `Gem::InstallError` occurs because `ruby_dep requires Ruby version >= 2.2.5, ~> 2.2.`.

b264: It appears this is the breakage https://github.com/guard/listen/commit/c65339c931c453848ed637b15edba5a064065d05

b264: https://github.com/guard/listen/blob/master/listen.gemspec#L37

b264: So these failures are intentional then. https://github.com/guard/listen/pull/378

ctran:
@akito19 Do you know how I can tell semaphoreci to clear the cache?

akito19:
@ctran Yes I want to know to clear cache.
And I'm sorry for dirty committing...

ctran:
That's ok, I'll squash these into a single commit upon merge.

ctran:
But I'd prefer you do a rebase before sending in the PR :-)

PR #405, opened 2016-10-07
Fixed Posgresql index tables with `public`.

ryanfox1985:
Hello,

I have a issue with my Postgresql database and the index annotations. 
In some models I added the table_name_prefix = 'public.' because I'm using different schemas and the annotation doesn't work because is looking for 'public.table_name' and doesn't exist.

Regards.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ group :development, :test do
   gem 'rubocop', '~> 0.37.2', require: false unless RUBY_VERSION =~ /^1.8/
   gem 'coveralls'
   gem 'codeclimate-test-reporter'
+  gem 'ruby_dep', '1.3.1'
 
   platforms :mri, :mingw do
     gem 'pry', require: false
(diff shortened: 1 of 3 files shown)


ctran:
I would like to see a test case added for this if you don't mind.

ryanfox1985:
Hi @ctran,

I will describe my case:

We are using `Postgresql` that supports to have multiples schemas (like `public` and `data_warehouse`).
Both schemas are sync (in structure) but we have some models using the `table_name_prefix` pointing to the `public` schema (like `Users`) because the data were the same. 

Let me add Rspec test to this part. 

Regards.

ctran:
@ryanfox1985 Did you have a chance to add a test?

ryanfox1985:
@ctran It's not easy to make a test xD. 
It will be the first that test the indexes.

ryanfox1985:
@ctran it's ready.

PR #418, opened 2016-12-12
Valid for rake earlier than v 12.1

BryanH:
Current Rake is v 12.0.0 and Gem won't install

### DIFF
--- a/annotate.gemspec
+++ b/annotate.gemspec
@@ -26,10 +26,10 @@ Gem::Specification.new do |s|
     s.specification_version = 4
 
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
-      s.add_runtime_dependency(%q<rake>, [">= 10.4", "< 12.0"])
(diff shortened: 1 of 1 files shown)


annotate.gemspec
```diff
@@ -26,10 +26,10 @@ Gem::Specification.new do |s|
     s.specification_version = 4
 
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
-      s.add_runtime_dependency(%q<rake>, [">= 10.4", "< 12.0"])
+      s.add_runtime_dependency(%q<rake>, [">= 10.4", "< 12.1"])
```
b264: Just curious, why not `"< 13.0"` ?

BryanH: 12 -> 13 should break a ton of stuff if they're using semantic versioning. I tested it with 12, but I can't be sure 13 will work out of the box until I get my hands on that version and can test.

b264: I mean less than 13 like `"< 13"`  but this merged already anyway :+1:

PR #420, opened 2016-12-13
Moved rubocop to outside.

ryanfox1985:
Hi, 

I moved Rubocop to the CI, it's better not runs Rubocop in the tests.
The best will be run in two pipelines.

Regards.


ctran:
Thanks!!!

PR #427, opened 2016-12-16
Replaced active-admin as a flag.

### LINKED ISSUE #363: "Annotate all activeadmin models" option is missing
It is only mentioned in the README.rdoc

ryanfox1985:
--active-admin replaced as a flag.
closes #363.


ryanfox1985:
included in PR #429

PR #432, opened 2017-01-03
Fix bug, the max array can contains a nil values.

ryanfox1985:
Apply compact to max array for filter nil values: https://github.com/ctran/annotate_models/blob/develop/lib/annotate/annotate_routes.rb#L44

### DIFF
--- a/lib/annotate/annotate_routes.rb
+++ b/lib/annotate/annotate_routes.rb
@@ -41,9 +41,10 @@ module AnnotateRoutes
       return out if routes_map.size.zero?
 
       maxs = [HEADER_ROW.map(&:size)] + routes_map[1..-1].map { |line| line.split.map(&:size) }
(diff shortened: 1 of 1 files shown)


lib/annotate/annotate_routes.rb
```diff
@@ -41,9 +41,10 @@ def header(options = {})
       return out if routes_map.size.zero?
 
       maxs = [HEADER_ROW.map(&:size)] + routes_map[1..-1].map { |line| line.split.map(&:size) }
-      max = maxs.map(&:max).max
 
       if options[:format_markdown]
+        max = maxs.map(&:max).compact.max
+
```
ctran: Is this var only used for markdown?

ryanfox1985:
Yes this variable is only used to generate the markdown.

Besides, the max array can contains null values is for this reason that I append compact.

Issue #380, opened 2016-05-27
"Custom" postgres database indexes not shown


micred:
We have many case-insensitive indexes in postgresql (as in the second example of [pg documentation](https://www.postgresql.org/docs/9.1/static/sql-createindex.html#AEN66303)).

These index are not printed by annotate.

How to replicate: 

```
class AddNameIndexToFilm < ActiveRecord::Migration
  def change
    execute 'CREATE INDEX index_films_on_lower_title ON films ((lower(title)));'
end
```

uberllama:
Funny, I was just about to post something similar, except in my case I want to see index types (btree, gin, etc).

ctran:
Is this similar to #377?

PR #442, opened 2017-01-30
Do not check if an index includes another column if the name is a String, instead of an Array.

adamgaynor:
## Issues Fixed
https://github.com/ctran/annotate_models/issues/380

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -103,7 +103,7 @@ Lint/UselessAccessModifier:
 
 # Offense count: 17
 Metrics/AbcSize:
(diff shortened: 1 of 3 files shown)


ctran:
Thanks for helping out but this PR causes CI failures.  Can you take a look?

adamgaynor:
@ctran It looks like what was happening is the complexity limitations set for Rubocop in your testing framework were pushed over the limit when I added an `if` statement in the PR. 
I've increased the complexity limits in the necessary places to make the tests pass, and also written an rspec test to ensure the change works.

PR #447, opened 2017-02-17
Update gemspec and Rakefile

dankimio:
- Moved dependencies from Gemfile to gemspec
- Cleaned up the gemspec
- Updated .gitignore
- Make VERSION a constant

Cleaned up Rakefile, not sure how those tasks were used, some feedback would be great.


ctran:
What's the goal of this?  Why do you want to move dependencies into gemspec?

dankimio:
@ctran Less code and the same functionality. Becomes easier to read in some places. Also less repetition (for example, Rakefile). I'll split up the PR soon.

Issue #340, opened 2016-02-02
Does not annotate factories after DB migration


heaven:
Here's my config:

``` ruby
Annotate.set_defaults(
  'routes'                  => 'false',
  'position_in_routes'      => 'after',
  'position_in_class'       => 'after',
  'position_in_test'        => 'after',
  'position_in_fixture'     => 'after',
  'position_in_factory'     => 'after',
  'position_in_serializer'  => 'after',
  'show_foreign_keys'       => 'true',
  'show_indexes'            => 'true',
  'simple_indexes'          => 'false',
  'model_dir'               => 'app/models',
  'root_dir'                => '',
  'include_version'         => 'false',
  'require'                 => '',
  'exclude_tests'           => 'false',
  'exclude_fixtures'        => 'false',
  'exclude_factories'       => 'false',
  'exclude_serializers'     => 'false',
  'exclude_scaffolds'       => 'false',
  'exclude_controllers'     => 'true',
  'exclude_helpers'         => 'true',
  'ignore_model_sub_dir'    => 'false',
  'ignore_columns'          => nil,
  'ignore_unknown_models'   => 'false',
  'hide_limit_column_types' => 'integer,boolean',
  'skip_on_db_migrate'      => 'false',
  'format_bare'             => 'true',
  'format_rdoc'             => 'false',
  'format_markdown'         => 'false',
  'sort'                    => 'false',
  'force'                   => 'false',
  'trace'                   => 'false',
  'wrapper_open'            => nil,
  'wrapper_close'           => nil,
)
end
```

After running migration it updates annotations in models, but not in factories, then I have to run `bundle exec annotate` to annotate factories:

```
➜  clarion-core git:(master) ✗ rake db:migrate
== 20160202113832 AddMaxDailyCountToReminder: migrating =======================
-- add_column(:reminders, :max_daily_count, :integer, {:default=>1, :null=>false})
   -> 0.0498s
== 20160202113832 AddMaxDailyCountToReminder: migrated (0.0499s) ==============

Annotated (1): app/models/reminder.rb

➜  clarion-core git:(master) ✗ bundle exec annotate
Annotated (1): app/models/reminder.rb

➜  clarion-core git:(master) ✗ bundle exec annotate
Model files unchanged.
```

vraravam:
This not only skips annotating factory_girl factories, but also skips rspec files. But, one interesting thing I noticed is that the `annotate` executable DOES annotate all such files while the rake task does not

[...]

PR #448, opened 2017-03-02
Prevent root_dir option from being empty

lime:
This is a quick fix for #340.

When setting root_dir to an empty string (as is done in the default rake task config) this line previously caused root_dir to become an empty array.

This in turn caused factories, specs etc not to be annotated.

This is just a quick band-aid on a larger problem. We have option parsing spread out over many different places, with slight mismatches like this in the assumptions made.

### DIFF
--- a/lib/tasks/annotate_models.rake
+++ b/lib/tasks/annotate_models.rake
@@ -21,7 +21,7 @@ task annotate_models: :environment do
   options[:show_indexes] = Annotate.true?(ENV['show_indexes'])
   options[:simple_indexes] = Annotate.true?(ENV['simple_indexes'])
   options[:model_dir] = ENV['model_dir'] ? ENV['model_dir'].split(',') : ['app/models']
-  options[:root_dir] = ENV['root_dir'] ? ENV['root_dir'].split(',') : ['']
+  options[:root_dir] = ENV['root_dir'] && !ENV['root_dir'].empty? ? ENV['root_dir'].split(',') : ['']
   options[:include_version] = Annotate.true?(ENV['include_version'])
   options[:require] = ENV['require'] ? ENV['require'].split(',') : []
   options[:exclude_tests] = Annotate.true?(ENV['exclude_tests'])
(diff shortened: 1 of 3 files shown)


lime:
I would probably approach the underlying issue by creating a `Annotate::Options` class, which could encapsulate all the quirks of our available options.

Then each place that uses these options (`bin/annotate`, rake tasks, direct invocation from code) wouldn't have to encode the same fallbacks & defaults, reducing the risk for bugs like this one.

I most likely won't have time to attempt that kind of overhaul, but maybe someone else would like to give it a try?

lib/tasks/annotate_models.rake
```diff
@@ -21,7 +21,7 @@ task annotate_models: :environment do
   options[:show_indexes] = Annotate.true?(ENV['show_indexes'])
   options[:simple_indexes] = Annotate.true?(ENV['simple_indexes'])
   options[:model_dir] = ENV['model_dir'] ? ENV['model_dir'].split(',') : ['app/models']
-  options[:root_dir] = ENV['root_dir'] ? ENV['root_dir'].split(',') : ['']
+  options[:root_dir] = ENV['root_dir'] && !ENV['root_dir'].empty? ? ENV['root_dir'].split(',') : ['']
```
ctran: I wonder if we should fix it where options[:root_dir] being used, make it work whether it's [], [''], or ['', '',...].

lime: 100% yes, we should. :)

I wasn't sure at first of how large the surface of `root_dir` was, so I wanted to start off with this simple fix. It seems though like this could really easily be taken care of inside `AnnotateModels.root_dir`.

Would you like me to give that a shot?

ctran: Please and thank you.

lime:
There, I changed the `root_dir` method to return `['']` for all blank values, split comma-separated strings and leave other values (i.e. arrays) untouched. How does that look?

lime:
> However, can I request that you move this logic into the setter (writer) so it's done once?

I guess, although personally I think I would keep it in the getter. :)

Should `model_dir` be changed then too, to keep some consistency between the two?

lime:
As an aside, how do you feel about the idea I mentioned earlier about moving all of this logic into `Annotate::Options` or something similar? If we know you are interested it might be more likely that someone takes a stab at it. :)

lime:
> However, can I request that you move this logic into the setter (writer) so it's done once?

I went to do this and then realized why it's problematic: with the getter you'll get `['']` even when you never set the value, while with the setter you run the risk of getting `nil` by default.

I would suggest we keep the current approach.

PR #454, opened 2017-04-27
Improve handling of default values for serialized attributes

johncarney:
If a serialized attribute has a default value, the annotation can include the object id of the default value, which changes every time `annotate` is run,

This PR encodes default values using the column's coder where available.

# Notes

There is another PR (#381) that attempts to solve this problem, but I think this solution is more robust and does not alter existing behaviour.

# TODO

- [ ] Check that it will work for supported versions of Rails
- [ ] Add tests


ctran:
Please make sure the tests passed on Travis.

johncarney:
Rebased on the wrong branch >_<

PR #455, opened 2017-04-29
Modify private class methods

yhirano55:
I found wrong definitions.

This does not work:

```
[1] pry(main)> module Foo
[1] pry(main)*   private
[1] pry(main)*   def self.say
[1] pry(main)*     __method__
[1] pry(main)*   end
[1] pry(main)* end
=> :say
[2] pry(main)> Foo.say
=> :say
```

We should define `private_class_method` or define inside its singleton class block:

```
[3] pry(main)> module Bar
[3] pry(main)*   def self.say
[3] pry(main)*     __method__
[3] pry(main)*   end
[3] pry(main)*   private_class_method :say
[3] pry(main)* end
=> Bar
[4] pry(main)> Bar.say
NoMethodError: private method `say' called for Bar:Module
from (pry):14:in `__pry__'
```

```
[5] pry(main)> module Baz
[5] pry(main)*   class << self
[5] pry(main)*     private
[5] pry(main)*     def say
[5] pry(main)*       __method__
[5] pry(main)*     end
[5] pry(main)*   end
[5] pry(main)* end
=> :say
[6] pry(main)> Baz.say
NoMethodError: private method `say' called for Baz:Module
from (pry):23:in `__pry__'
```

### DIFF
--- a/lib/annotate/annotate_routes.rb
+++ b/lib/annotate/annotate_routes.rb
@@ -74,125 +74,125 @@ module AnnotateRoutes
         puts "Removed annotations from #{routes_file}."
       end
     end
-  end
 
-  private
+    private
 
-  def self.app_routes_map(options)
-    routes_map = `rake routes`.split(/\n/, -1)
+    def app_routes_map(options)
+      routes_map = `rake routes`.split(/\n/, -1)
 
-    # In old versions of Rake, the first line of output was the cwd.  Not so
-    # much in newer ones.  We ditch that line if it exists, and if not, we
-    # keep the line around.
-    routes_map.shift if routes_map.first =~ /^\(in \//
+      # In old versions of Rake, the first line of output was the cwd.  Not so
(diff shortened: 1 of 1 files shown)


ctran:
Could you explain a bit the reasons for this change?

yhirano55:
Thanks for your reviewing.

[The current code](https://github.com/ctran/annotate_models/blob/develop/lib/annotate/annotate_routes.rb#L79) expects `private`, but as a behavior it does not become `private` ([*](https://github.com/ctran/annotate_models/pull/455#issue-225242942)). So we had better to remove `private` keyword in `lib/annotate/annotate_routes.rb`, or to modify correct syntax🤓

yhirano55:
@ctran Please re-review, or close this request if you don't need (It's no problem for me)

ctran:
@yhirano55 Thanks for the education.  You're correct that "private" doesn't work in this case.  I'll remove the private keyword since it's less intrusive.

PR #471, opened 2017-06-02
Remove "private" since it's not valid in context

ctran:
Pointed out by @yhirano55 in #455, the private keyword is not valid in this context.


yhirano55:
+1

PR #475, opened 2017-06-05
Fix show_complete_foreign_keys setting

steveklebanoff:
https://github.com/ctran/annotate_models/pull/458 introduced a `show_complete_foreign_keys` setting that defaults to `false`.  But, setting it to `true` didn't actually work, because it was excluded from `FLAG_OPTIONS`


steveklebanoff:
@thedanielhanke ^

thedanielhanke:
nice, thanks!

steveklebanoff:
np @thedanielhanke .  also, just realized this option should also be added to https://github.com/ctran/annotate_models/blob/develop/bin/annotate .  wanna take a stab at it?

thedanielhanke:
@steveklebanoff will stab it to death :+1: :)

Issue #476, opened 2017-06-08
Add show_complete_foreign_keys to bin


thedanielhanke:
Currently, there is no way to add the show_complete_foreign_keys flag to the cli, which i forgot. see #475

PR follows.

PR #477, opened 2017-06-08
#476 add missing complete-foreign-keys option to bin

thedanielhanke:
adds the show_complete_foreign_keys flag to the cli, which i forgot. see #475.

i named it `--ck` in short, to have it aligned under `-k` hence is why i also set the `show_foreign_keys` option positive.

had to increase the `Metrics/BlockLength` a bit to make rubocop pass. Since `bin/annotate` is a good candidate for even more lines, i`d recommend to keep smaller blocks and exclude the file from this check.
However, this is a entirely different discussion which is why i did the increment only.

Did not find a spec for the bin options itself. is it missing or is it just me?

also, sorry all for the wrong branch name after an creating a useless ticket, bad memory.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -108,7 +108,7 @@ Metrics/AbcSize:
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/BlockLength:
-  Max: 134
+  Max: 140
 
 # Offense count: 2
 Metrics/BlockNesting:
--- a/README.rdoc
+++ b/README.rdoc
@@ -184,6 +184,8 @@ you can do so with a simple environment variable, instead of editing the
         -m, --show-migration             Include the migration version number in the annotation
         -i, --show-indexes               List the table's database indexes in the annotation
         -k, --show-foreign-keys          List the table's foreign key constraints in the annotation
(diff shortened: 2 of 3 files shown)


thedanielhanke:
@steveklebanoff ^

steveklebanoff:
image

Formatting looks a little funky on github, description should be on same line.  Besides that, looks good to me as long as @ctran is cool with the rubocop change

thedanielhanke:
@steveklebanoff the formatting is due to the length of the key. 
i think `complete-foreign-keysComplete...` will not look more decent :P

i could however, indent the entire right readme section a bit, or remove one of the two options. 
The optionparser would need to receive the same change:
```
$ ./bin/annotate --help | ag -B1 -A2 '\--ck' 
    -k, --show-foreign-keys          List the table's foreign key constraints in the annotation
        --ck, --complete-foreign-keys
                                     Complete foreign key names in the annotation
    -i, --show-indexes               List t....
```
which is all unrelated somehow. so i'd await @ctran`s opinion on this.
suggestions from you, @steveklebanoff ?

thedanielhanke:
@steveklebanoff I forgot the most important part: thanks for your review! :+1:

ctran:
Thanks much!

Issue #403, opened 2016-09-21
Ordered index doesn't appear with the order


gregeinfrank:
I created an index in `DESC` order in `postgresql`.  The index in `psql` looks like:

``` sql
Indexes:
    "index_id_and_created_at" btree (id, created_at DESC)
```

But the annotation in my model from `annotate_models` looks like:

``` ruby
# Indexes
#
#  index_id_and_created_at  (id,created_at)
```

Let me know if I'm missing something here.
Thanks.

PR #479, opened 2017-06-19
Annotate ordered indexes

sashabelozerov:
Solves #403 

Now if you have ordered indexes, order direction will be added next to the column name.

Example:

```
# Indexes
#
#  index_rails_02e851e3b7  (id)
#  index_rails_02e851e3b8  (name DESC,value)
```

Markdown is also supported.


ctran:
Thanks!

PR #422, opened 2016-12-14
Add more information for partial indices

indigoviolet:
- includes the `WHERE` clause for partials
- the index type if not `btree`

PR #482, opened 2017-07-02
Indexes: add WHERE and USING

sashabelozerov:
This pull request is inspired by other PR https://github.com/ctran/annotate_models/pull/422
It adds WHERE and USING values to the index info.
This is really useful feature, and since original PR contributor does not appear since Dec 2016, I decided to push it forward. Hope this is fine.

USING is added only if index is not `btree`.

Difference from #422:
1) Markdown support
2) Refactored code
3) Added missing specs

Example:

```
# Indexes
#
#  index_rails_02e851e3b7  (id)
#  index_rails_02e851e3b8  (firstname,surname) WHERE city is not null USING hash
```


ctran:
Thanks!

PR #485, opened 2017-07-11
annotate --routes modifies only routes.rb

Hitabis:
This change is a proposed solution to #357 and an alternative to the already proposed solution #361.
In #361 it is needed to call `annotate --routes --ignore-models` to achieve the same as in this change with only `annotate --routes`.

### DIFF
--- a/bin/annotate
+++ b/bin/annotate
@@ -202,7 +202,7 @@ end.parse!
 options = Annotate.setup_options(
   is_rake: ENV['is_rake'] && !ENV['is_rake'].empty?
 )
-Annotate.eager_load(options)
+Annotate.eager_load(options) if Annotate.include_models?
 
 AnnotateModels.send(target_action, options) if Annotate.include_models?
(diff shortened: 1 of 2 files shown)


ctran:
Does this mean I can't do both routes and models now?

Hitabis:
Yes, that's what I would expect from:
- reading the README
https://github.com/ctran/annotate_models/blame/330f8bdb8dbcf6ea62b8f428d3cdca540dc4deb5/README.rdoc#L95-L97
- the cli help
https://github.com/ctran/annotate_models/blob/134b833ec90830b95628defcc162536d3670b9fa/bin/annotate#L89

ctran:
I misunderstood the previous behavior.  I'll merge this in.  Thanks.

Hitabis:
@ctran Any updates on the v2.7.3 release?

jjercx:
This means if I `'routes' => 'true'` in `auto_annotate_models.rake`, it won't work with my models?

PR #494, opened 2017-07-25
Fix README

yhirano55:
I've added option `require: false` on Usage. Because this gem doesn't need autorequiring. And I've refactored tiny things. Thanks.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -55,11 +55,11 @@ Also, if you pass the -r option, it'll annotate routes.rb with the output of
 
 Into Gemfile from rubygems.org:
 
-    gem 'annotate'
+    gem 'annotate', require: false
 
 Into Gemfile from Github:
 
-    gem 'annotate', git: 'https://github.com/ctran/annotate_models.git'
+    gem 'annotate', github: 'ctran/annotate_models', require: false
 
 Into environment gems from rubygems.org:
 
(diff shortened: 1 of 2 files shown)


README.rdoc
```diff
@@ -55,11 +55,11 @@ Also, if you pass the -r option, it'll annotate routes.rb with the output of
 
 Into Gemfile from rubygems.org:
 
-    gem 'annotate'
+    gem 'annotate', require: false
 
 Into Gemfile from Github:
 
-    gem 'annotate', git: 'https://github.com/ctran/annotate_models.git'
+    gem 'annotate', github: 'ctran/annotate_models', require: false
```
b264: This will cause an non-secure warning, right?

yhirano55: This block is which was declared on the top of Gemfile:

```ruby
git_source(:github) {|repo_name| "https://github.com/#{repo_name}.git" }
```

Its a workaround for a bug in Bundler which can cause sources from github to loaded via HTTP and not HTTPS - which makes it vulnerable to man in the middle attacks. We'll usually not remove it.

ctran: Thanks for pointing this out to me.  The short form doesn't seem to gain us much and has a security downside.  I would rather use the longer form in this case.

yhirano55: Oh, I see. I'll fix it, soon.

ctran: @yhirano55 don't worry about it, I already started #496.

PR #496, opened 2017-07-26
Use long form instead github due to non-https url

ctran:


README.rdoc
```diff
@@ -59,7 +59,7 @@ Into Gemfile from rubygems.org:
 
 Into Gemfile from Github:
 
-    gem 'annotate', github: 'ctran/annotate_models', require: false
+    gem 'annotate', git: 'https://github.com/ctran/annotate_models.git', require: false
```
yhirano55: 👍

PR #513, opened 2017-09-19
Fix annotations for rails 5.1.4

anilmaurya:
- annotate command generate annotation like
  name        :#<ArJdbc::MySQL: default(""), not null
  path        :#<ArJdbc::MySQL:
 ...

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -249,7 +249,8 @@ module AnnotateModels
       cols = cols.sort_by(&:name) if options[:sort]
       cols = classified_sort(cols) if options[:classified_sort]
       cols.each do |col|
-        col_type = ((col.type && col.type.type) || (col.sql_type && col.sql_type.to_s))
+        col_type = (col.type || col.sql_type)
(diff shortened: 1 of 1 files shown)


lib/annotate/annotate_models.rb
```diff
@@ -249,7 +249,7 @@ def get_schema_info(klass, header, options = {})
       cols = cols.sort_by(&:name) if options[:sort]
       cols = classified_sort(cols) if options[:classified_sort]
       cols.each do |col|
-        col_type = (col.type || col.sql_type).to_s
+        col_type = ((col.type && col.type.type) || (col.sql_type && col.sql_type.to_s))
```
b264: how about
```ruby
col_type = (col.type || col.sql_type)
col_type = col_type.type if col_type.respond_to?(:type)
col_type= col_type.to_s
```

Because the one above doesn't look like it will continue working for older Rails versions

anilmaurya:
@b264 Updated code to support earlier version of rails

drwl:
Hi @anilmaurya I'm in the process of cleaning up old branches. It's been a few years since this PR was first put up and there may have been code changes since then. If you're still interested in getting the change in feel free to reopen this.

PR #515, opened 2017-10-11
annotate bigint correctly

willnet:
Bigint columns display integer in annotation. Because  `ActiveRecord::ConnectionAdapters::Column#type` returns `:integer` if its column type is bigint.

`#bigint?` is provided to distinguish them so I changed to use it.


ctran:
Thanks!

Issue #433, opened 2017-01-09
[Feature request] Support database comments


okuramasafumi:
Rails 5 supports comments in database. It would be convenient if they appear in annotations.

deepj:
Rails 5 supporting adding comments in migrations

http://blog.bigbinary.com/2016/06/21/rails-5-supports-adding-comments-migrations.html

hemju:
This would be a great feature. Anything we can help you to implement it?

gambala:
PR was merged, but feature didn't work for me :(

Ruby 2.4.1, Rails and ActiveRecord 5.1.1, annotate 2.7.2.

Used both `annotate` and `rake annotate_models`. Rake task was generated by `annotate:install` with this option in generated file:

```
'with_comment'              => true
```

okuramasafumi:
Thank you for your work @motot and @dslh!
It seems the final working version is not released yet, so I cannot close this issue.
I'm waiting for the new release, @ctran!

Adeynack:
@ctran : You closed the issue on Feb. 14th. However, the last release I see on GH dates 2018. Is this feature here bundled in this last release (2.7.4) or not yet? If not yet, when will it be? My team wants to use that feature.

PR #518, opened 2017-10-13
Make comment annotations work

dslh:
This is a follow up for issue #433. PR #459 implements the feature, but forgets to hook it up to either the rake task or the command line executable. I've added it to both ☺️

I've also fixed a bug where `with_comment` would fail if any columns of a table were uncommented. The feature will also now turn itself off if a table has no comments at all.

Rubocop complained about various complexity metrics on the `get_schema_info` method, so I've moved a couple of bits of code into private methods. After that I've rebuilt `.rubocop-todo.yml`, which is looking a bit better than the last time it was generated. The only cop that's gotten worse is `Metrics/BlockLength`, which has gone up due to the increased size of the `OptionParser.new` block in `bin/annotate` 😬 

Thanks very much for this gem, I hope this PR will be useful to others 🙂


ctran:
Thanks!!!

PR #528, opened 2017-10-27
Redirect error messages to current standard error stream

tonyta:
We had a script that ran a Rails migration. Since its output is noisy, we swallowed stdout. The exit code was `1` but there was no message because `annotate_models` puts its error messages to stdout instead of stderr.

This PR redirects error messages to the current standard error stream `$stderr`.

**Before**
```
$ bundle exec rake annotate_models >/dev/null
$ echo $?
1
```

**After**
```
$ bundle exec rake annotate_models >/dev/null
No models found in directory 'app/models'.
Either specify models on the command line, or use the --model-dir option.
Call 'annotate --help' for more info.
$ echo $?
1
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -28,7 +28,7 @@ require 'mg'
 begin
   MG.new('annotate.gemspec')
 rescue Exception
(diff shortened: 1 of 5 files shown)


ctran:
Thanks!

Issue #413, opened 2016-11-24
Error for model in subdirectory


axhamre:
I'm on Rails 5.0.

Annotate has worked fine for me until I put my models in different subdirectories under the directory 'models'.

Now I get the following notive (which is incorrect) for two of my models (i.e. not all) when running `annotate -p bottom`:

    Unable to annotate app/models/physical/meeting_link.rb: You tried to define an enum named "availability" on the model "MeetingLink", but this will generate a instance method "default_availability?", which is already defined by another enum.

Tybot204:
This happens to me as well on Rails 4.2.7. All of my enum attributes give a similar set of errors when the model is placed inside of a sub directory. This error prevents the model from being annotated, and is not just a warning message.
```
Unable to annotate app/models/tms/fee.rb: You tried to define an enum named "fee_type" on the model "Fee", but this will generate a instance method "fee_type=", which is already defined by another enum.
```

davidlbean:
+1

davidlbean:
@hundsim @Tybot204 - so I was having the same problem and I dug into things to see if I could produce a solution...  I found that the error we're all getting is from ActiveRecord itself when Annotate loads the model file.  In my case, it was because I had defined a model in a subdirectory without adding a matching namespace via a module.  Specifically:  
'''
I have a directory structure like this:
-- app
 -- models
   -- my_class_a.rb
   -- my_class_b.rb
   -- summary_review
       -- my_nested_class_with_enum.rb
       -- my_nested_class_without_enum.rb
'''

Initially, my_nested_class_with_enum.rb defined:
'''
   class MyNestedClassWithEnum
     ...
   end       
'''

But, once I defined it with a namespace like this:
'''
   module SummaryReview
      class MyNestedClassWithEnum
         ...
       end
   end
'''
everything worked fine. 

Hope that helps.

- David

Tybot204:
@davidlbean Ah yes that looks like it might be my problem as well.
My directory structure:
```
-- app
 -- models
   -- my_class_a.rb
   -- my_class_b.rb
   -- directory_in_models
       -- my_nested_class_with_enum.rb
       -- my_nested_class_without_enum.rb
```
I was then auto loading these models so Rails treated them the same as if they were not nested:
```ruby
## Inside application.rb
config.autoload_paths += Dir[Rails.root.join('app', 'models', 'directory_in_models', '**/')]
```

[...]

PR #533, opened 2017-12-01
add spec for #413

manuelmeurer:
This is the setup that should trigger #413.
Since I don't use RVM and a lot of tests seem to depend on it, let's get Travis going! 😄


ctran:
Great!  I'll take a look.

PR #535, opened 2017-12-01
[Fix #281, #413]: Try to get proper loaded model when using Rails eager_load_paths

### LINKED ISSUE #281: 'already initialized constant' warnings when models are in additional configured auto load roots
###### to repro:
1. `git clone git@github.com:bughit/annotate_test.git`
2.  `rake db:migrate`

###### how to configure additional auto load roots:
`config.eager_load_paths.concat Dir["#{config.root}/app/models/additional_roots/**/"]`

0x01f7:


ctran:
Could you add the spec mentioned in #535 so we can see that it did fix the problem?

0x01f7:
@ctran added a test case

ps: the integration test is temporarily ignored in Travis, and i cannot make it work on my personal notebook too
pps: I use a repo baed on #281 to test it

PR #552, opened 2018-04-18
Fix rubocop violation (extra line added before model definition)

a14m:
Remove the extra `\n` inserted before the class causing the diff to look something like

```diff
+# **`sprite_length`**        | `integer`          |
+#
+
+
 class Video < ApplicationRecord
```
causing rubocop to complain `C: Layout/EmptyLines: Extra blank line detected.`

solution: Remove the extra line so that the diff would look like:
```diff
+# **`sprite_length`**        | `integer`          |
+#
+
 class Video < ApplicationRecord
```

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -535,7 +535,7 @@ module AnnotateModels
             new_content = if %w(after bottom).include?(options[position].to_s)
                             magic_comments_block + (old_content.rstrip + "\n\n" + wrapped_info_block)
                           else
-                            magic_comments_block + wrapped_info_block + "\n" + old_content
+                            magic_comments_block + wrapped_info_block + old_content
                           end
           end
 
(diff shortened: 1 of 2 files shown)


kreintjes:
@a14m I just tried your fork and it works nicely if you don't have magic comments. However unfortunately, things go wrong when there are magic comments on top of the file (e.g. `frozen_string_literal: true`). It then changes this:

```
# frozen_string_literal: true

class Tag < ApplicationRecord
```

Into this:

```
# frozen_string_literal: true
# == Schema Information
#
# Table name: tags
#
#  id         :integer          not null, primary key
# ...
#


class Tag < ApplicationRecord
```

So it seems the annotations are inserted in the wrong place (directly after the magic comment, while they should be directly before the class). This results in two RuboCop violations:
- "Layout/EmptyLineAfterMagicComment: Add an empty line after magic comments."
- "Layout/EmptyLines: Extra blank line detected."

Any chance you could take a look at this? This might be fixed already in the original repo. Could you rebase your fork on the upstream?

PR #553, opened 2018-04-19
Use ANNOTATE_SKIP_ON_DB_MIGRATE ENV variable

trevorrjohn:
I think this makes it clearer about what we are actually setting with the environment variable.


ctran:
Thanks!

PR #560, opened 2018-05-01
Fix with_comment format in annotate_models_rake

yhirano55:
### Summary

* :with_comment option is flag.
* It should use `true?` instead of `fallback`.

Now with_comment option is enforcely overwritten in annotate_models.rake. So I've fixed it.


yhirano55:
@ctran When is it going to release the new version which includes this patch?

ctran:
@yhirano55 it's published - https://rubygems.org/gems/annotate/versions/2.7.4

yhirano55:
Thank you so much!

PR #561, opened 2018-05-01
Use string instead of boolean

yhirano55:
### Summary

:with_comment option should use string instead of boolean as same as others.


yhirano55:
@ctran How about this pr? I feel it's so tiny change but it's going to be comfortable for almost of all developers which love this library.

yhirano55:
@ctran Thank you so much :)

PR #573, opened 2018-09-06
Keep same surrounding whitespace when re-annotating a file

CyborgMaster:
This fixes #555, fixes #472, and possibly more.

This is currently built on top of #571, because it hits the same lines of code.  If for some reason you want to merge this one and not #571, I'll be happy to refactor.

PR #574, opened 2018-09-06
Refactor and simplify returns

CyborgMaster:
Refactor and simplify returns for `annotate_one_file`.   This simplifies some big if/else blocks that were wrapping significant portions of this method.  No logic is changed.

This is best viewed hiding whitespace changes (under the "Diff settings" button).

This is currently built on top of #573, because it hits the same lines of code (I split it out because the whitespace changes make it hit every line in the function). If for some reason you want to merge this one and not #573, I'll be happy to refactor.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -512,7 +512,7 @@ module AnnotateModels
       new_columns = new_header && new_header.scan(column_pattern).sort
 
       return false if old_columns == new_columns && !options[:force]
-      
+
       # Replace inline the old schema info with the new schema info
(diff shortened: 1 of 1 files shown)


CyborgMaster:
After you get 573 merged, I'll rebase this one too and hopefully that will make the diff simpler.

CyborgMaster:
Rebased!  If you hide the whitespace changes now, the diff is quite short.

ctran:
Thanks!!!

CyborgMaster:
You're welcome.  Thanks for getting these merged so fast.  I don't think I've seen such a quick turn around on an OSS project before.  Much appreciated.

PR #575, opened 2018-09-08
Fix alignment of multi-byte fullwidth character comments

oieioi:
This fixes alignment of Japanese, Korean and Chinese fullwidth character comments.

The displayed widths of multi-byte fullwidth characters are generally
twice as large as the ASCII characters, but String#size returns only the number of characters.
So if the column comment contains fullwidth multibyte characters, the alignment is broken.

Monospaced font-family supporting multi-byte isn't used in GitHub diff view, you can confirm the fixed alignment with `font-family: "Osaka-mono", "MS Gothic";`.

cf. https://gyazo.com/d878a077b7494f07bcc45ecf7c0adb62


ctran:
Thanks!

PR #584, opened 2018-11-09
Do not add whitespace after magic comments when annotating at position after/bottom

tjwp:
When position is set to after or bottom the current behavior added an extra newline following magic comments. 

If the file is being annotated at position after/bottom there should be no need to make modifications after the magic comments. (Any style/formatting there is not the concern of this gem if the file is not being annotated there.)

This change fixes that behavior to only insert a newline if the file is being annotated at position before/top, i.e. between the magic comments and existing content.


ctran:
Thanks!!!

NikoRoberts:
Whitespace is still added when there is a module wrapping the class

PR #590, opened 2019-01-07
Allow ignore_unknown_models to be overridden by an environment variable

benrbrook:
Allow ignore_unknown_models to be overridden by an environment variable. This allows people to silence errors from non-AR models present in the `models` folder.


ctran:
I'm not sure why travis is failing.  Restarting the build to verify.

ctran:
Thanks!

PR #598, opened 2019-02-02
Make methods private in AnnotateRoutes

nard-tech:
First I tried to extend `AnnotateModels`, but I found several problems in `AnnotateModels` and `AnnotateRoutes`.

In `AnnotateRoutes`, there were many public methods that is not used in other classes or modules.

Before my extension, I made these methods private and sort them in order of appearance.

All tests were passed.

Please trace each commits carefully. I didn't nothing special. The commits consist of coordinating code.

## Note

As [Code Climate](https://codeclimate.com/github/ctran/annotate_models) shows, this gem is hard to maintain. 

So I'd like to refactor this gem in the near future.

### DIFF
--- a/lib/annotate/annotate_routes.rb
+++ b/lib/annotate/annotate_routes.rb
@@ -148,6 +148,21 @@ module AnnotateRoutes
       # TODO: below...
       content
     end
+
+    # @param [String, Array<String>]
+    def rewrite_contents(existing_text, new_content)
+      # Make sure we end on a trailing newline.
+      new_content << '' unless new_content.last == ''
+      new_text = new_content.join("\n")
+
+      if existing_text == new_text
+        puts "#{routes_file} unchanged."
+        false
+      else
+        File.open(routes_file, 'wb') { |f| f.puts(new_text) }
+        true
+      end
+    end
   end
 
   def self.magic_comment_matcher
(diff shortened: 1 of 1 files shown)


lib/annotate/annotate_routes.rb
```diff
@@ -35,41 +35,6 @@ def content(line, maxs, options = {})
...
-      magic_comments_map, routes_map = extract_magic_comments_from_array(routes_map)
-
-      out = []
-
-      magic_comments_map.each do |magic_comment|
-        out << magic_comment
-      end
-      out << '' if magic_comments_map.any?
-
-      out += ["# #{options[:wrapper_open]}"] if options[:wrapper_open]
```
drwl: I know you didn't make this change, but out of curiosity do you know if there's a reason behind using `<<` versus `+= []` for appending to an array?

drwl:
👍 Changes look good to me. Looks a lot cleaner now that class methods are together in the singleton class rather than defined as `self._`.

lib/annotate/annotate_routes.rb
ctran: There's a semantic difference between them:
> https://stackoverflow.com/questions/4684446/why-is-the-shovel-operator-preferred-over-plus-equals-when-building-a

drwl: That makes sense but this is mutating an array I believe. Line 43 is `out = []`

drwl: Just tested it in Pry looks like the same applies regardless of if it's a String or an Array. Functionally it looks like one can be substituted over the other so would it makes sense to change the cases so they're the same?

drwl:
@ctran unsure if you need anything else but it should be good to merge in as it's not changing any functionality.

drwl:
Planning on landing this in the next few days and bumping the gem version to `3.0.0` to follow semantic versioning. Any thoughts @nard-tech or @ctran?

PR #603, opened 2019-02-23
Add frozen option

### LINKED ISSUE #317: Question: CI Support
We often forget to roll annotations. Is there a way to let annotate_models check if everything is annotated  and if that isn't true, throw an abort so our CI build fails.

Thnx upfront.

yokomotod:
fix #317 

Add `--frozen` option to abort instead of write annotation.
This is mainly for CI environment to detect missing annotation.
Similar motivation with `bundle install --frozen`.

works like:
```
$ rake db:migrate frozen=yes
..
annotate error. spec/models/user_spec.rb needs to be updated, but annotate was run with `--frozen`.
```

### DIFF
--- a/bin/annotate
+++ b/bin/annotate
@@ -22,6 +22,7 @@ has_set_position = {}
 target_action = :do_annotations
 positions = %w(before top after bottom)
 
+# rubocop:disable Metrics/BlockLength
 OptionParser.new do |opts|
   opts.banner = 'Usage: annotate [options] [model_file]*'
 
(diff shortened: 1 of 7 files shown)


bin/annotate
```diff
@@ -22,6 +22,7 @@ has_set_position = {}
 target_action = :do_annotations
 positions = %w(before top after bottom)
 
+# rubocop:disable Metrics/BlockLength
```
yokomotod: [note]

finally it reached to 
```
bin/annotate:25:1: C: Metrics/BlockLength: Block has too many lines. [145/142]
```
and I thought it's rational to suppress here IMO

percyhanna:
This PR looks great, and would be a great addition to an awesome gem. Just wondering if version 2.7.5 has been officially released yet? The link [in the changelog](https://github.com/ctran/annotate_models/commit/6a43890e333be85098ddbdfd71da4cf5747f5403#diff-00fc3968d7b2ccf1d9e755766e6a8854R2) 404's, and version 2.7.5 has not been published to RubyGems.

Thanks for the great work 👍

PR #623, opened 2019-05-30
Account for big string in models

rahul342:
I have this spatial type column using mysql's spatial adapter. It generates col_type to be `:geometry({:type=>"polygon", :srid=>0})`
which seems to be bigger than `bare_type_allowance`  making `padding` negative and causing exception. Checking for negative for such cases.


rahul342:
cc @ctran

PR #625, opened 2019-05-30
Fix serializers test directory files

morellan:
The current serializers test directory files assumed that the file ended in `_spec.rb` instead of `_test.rb`, that's why I wasn't getting my serializer tests annotated.


drwl:
@morellan is there any issue associated with this?

morellan:
I don’t think so, should I open one?

drwl:
If you wouldn't mind, doing so would add more context.

PR #626, opened 2019-05-30
allow annotate to work with Jets, at least for models

tongueroo:
Allows annotate to work with models for Jets. It won't work for Jets routes though.

Context: https://community.rubyonjets.com/t/using-annotate-with-jets/184

### DIFF
--- a/lib/annotate.rb
+++ b/lib/annotate.rb
@@ -143,7 +143,9 @@ module Annotate
     load_requires(options)
     require 'annotate/active_record_patch'
 
-    if defined?(Rails::Application)
+    if defined?(Jets)
+      Jets::Autoloaders.main.eager_load
+    elsif defined?(Rails::Application)
       if Rails.version.split('.').first.to_i < 3
         Rails.configuration.eager_load_paths.each do |load_path|
           matcher = /\A#{Regexp.escape(load_path)}(.*)\.rb\Z/


drwl:
@tongueroo could you add a bit more context onto what Jets is and how this would help?

tongueroo:
Sure. Jets is a Ruby Serverless Framework https://rubyonjets.com/ It supports ActiveRecord so that's why someone was wanting to use annotate with it. The changes only work for models, so understand if it doesn't suffice.

drwl:
I don't see why this gem couldn't support Jet. In its current implementation there aren't any tests or documentation on it in the README. Would it be possible to add either of these?

Also, what are your thoughts on adding a flag to invoke annotate for Jet models. Perhaps something like `--jet-models`?

cc @ctran if you have any objections

ctran:
> I don't see why this gem couldn't support Jet. In its current implementation there aren't any tests or documentation on it in the README. Would it be possible to add either of these?
> 
> Also, what are your thoughts on adding a flag to invoke annotate for Jet models. Perhaps something like `--jet-models`?
> 
> cc @ctran if you have any objections

Are we talking about annotating Rails models under the Jets framework?  If so, why would we need another flag?

drwl:
I believe the models used by Jets is the same type and kind used in Rails, that is ActiveRecord. 

With no objections I'm happy to merge it in @tongueroo. Could you add a test example with Jets? I plan to do some refactoring of the Gem and don't want to break behavior.

tongueroo:
> RE: Could you add a test example with Jets?

Sure. Will likely take some time to get to.

drwl:
Sure no problem @tongueroo, let me know when it's ready for review.

drwl:
@tongueroo Is this something you're still interested in getting merged in? 

I've added working integration tests. Should be straight forward to get a small Jets app using annotate, and then I would be comfortable merging in after.

tongueroo:
Unsure if will get to this 😔 Feel free to close. Can post a comment if end up getting to it.

drwl:
> Unsure if will get to this 😔 Feel free to close. Can post a comment if end up getting to it.

Sure no problem. Feel free to re-open or create a new PR.

PR #627, opened 2019-06-13
fix bundle exec annotate doesnt run on models, just routes

StecRichard:
Fix issue: bundle exec annotate doesn't run on models, just routes


drwl:
Thanks @StecRichard looks like this should fix https://github.com/ctran/annotate_models/issues/563.

Before merging it in, would it be possible to add some tests? I realize there isn't a spec for the rake tasks, so I can add that in if that'll help.

drwl:
@StecRichard gonna be closing this since it was fixed in #647.

PR #628, opened 2019-06-16
Update .travis.yml to unblock Ruby 2.2 builds

drwl:
This PR updates Ruby `2.2.X` to use `2.2.10`. It also unblocks Ruby 2.2 builds currently failing in CI. This happens because builds try updating to RubyGems 3 which does not support Ruby 2.2.


drwl:
cc @ctran just a fyi

PR #634, opened 2019-06-30
Suppress `warning: BigDecimal.new is deprecated`

wonda-tea-coffee:
It is as the subject.

Refer
ruby/bigdecimal@5337373


drwl:
Thanks for taking action and submitting a PR. Looks good to me so will be merging it in.

Currently I'm wrapping up some breaking changes so expect this to be out in the 3.0 release.

PR #635, opened 2019-07-01
Add tests for the CLI

drwl:
Currently there's no test coverage on CLI. This adds tests for a new class `Parser` which will replace the command line options currently in `bin/annotate`.

The technical direction I'm planning to go is to remove `ENV` variables completely and to have things passed into as arguments. Will be adding deprecation warnings in 3.0 and then deprecate `ENV` variables completely in 3.1. Deprecation warnings will be non-blocking in 3.0 when an `ENV` variable is set and an argument isn't passed in, and will becoming blocking in 3.1.


drwl:
@ctran I'll be merging this in

PR #637, opened 2019-07-11
Additional file patterns docs

ryanwjackson:
Add better documentation for new option.


README.rdoc
```diff
@@ -206,15 +207,35 @@ you can do so with a simple environment variable, instead of editing the
...
+
+CLI: +--additional_file_patterns+<br>
+Ruby: +:additional_file_patterns+
+
+Provide additional paths for the gem to annotate.  These paths can include globs.
+It is recommended to use absolute paths.  Here are some examples:
+
+
+- <code>/app/lib/decorates/%MODEL_NAME%/*.rb</code>
+- <code>/app/lib/forms/%PLURALIZED_MODEL_NAME%/**/*.rb</code>
```
ryanwjackson: @drwl I cannot figure out how to get `**` to escape correctly in rdoc rendering.  Any clue?  I've tried so many ways.

drwl: Try `&ast;`, I believe it should work.

- <code>/app/lib/decorates/%MODEL_NAME%/*.rb</code>
- <code>/app/lib/forms/%PLURALIZED_MODEL_NAME%/&ast;&ast;/*.rb</code>

Source: https://github.com/adam-p/markdown-here/issues/287#issuecomment-282429074

ryanwjackson: I actually can't replicate that.  Haha, I can't see the markdown for your comment so I can't copy it either.  Tried just using `&ast;` but that doesn't work either.  Mind just adding a commit?

PR #641, opened 2019-07-16
Refactor Parser

drwl:
Refactored Parser to isolate changes being made to ENV. This way we have an intermediate step where we know the environment variables being set.


KentBeck:
One step forward!

twwright:
Still having a look at the codebase for this gem for the first time, but it looks like line 59 of this commit may be causing issue #802 

https://github.com/ctran/annotate_models/blob/407ace641c637567372c98f32a6af10bbb1bc901/lib/annotate/parser.rb#L56-L60

PR #643, opened 2019-08-08
Set up CI with Azure Pipelines

ctran:


drwl:
@ctran are there any next steps for this PR?

ctran:
No need for this anymore since Github will have this builtin.  Would you have time looking into the new Github Actions that has builtin CI/CD?

PR #649, opened 2019-09-09
Fix --models option description

rnitta:
make compatible with README.md

### DIFF
--- a/lib/annotate/parser.rb
+++ b/lib/annotate/parser.rb
@@ -123,7 +123,7 @@ module Annotate
         env['routes'] = 'true'
       end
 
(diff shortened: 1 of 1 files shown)


drwl:
@rnitta thanks for this PR. Just curious what was the motivation for the change? Was the old description not helpful enough?

rnitta:
Is the old one correct?
In my understanding, --models option means "add schema annotation comments to rb files under app/models directory."

drwl:
That's a good point. It seems like it annotates `ActiveRecord` files that live in `app/models`. Thanks for the contribution @rnitta!

PR #650, opened 2019-09-13
Fix shifted when format_markdown option enabled and used non-ascii

hasimoto1009:
**before**
Screen Shot 2019-09-04 at 19 19 06

**after**
Screen Shot 2019-09-04 at 19 41 23


drwl:
@hasimoto1009 thanks for the PR! Would it be possible for you to add test(s) for this to show where it breaks and to prevent any future regressions?

hasimoto1009:
I'll try it. :)

PR #653, opened 2019-09-24
Refactor and tidy lib/annotate.rb

drwl:
Adds tests for `.include_routes?`, `.include_models?`, `.skip_on_migration?`. Also moves the `TRUE_RE` under the `Annotate::Constants` namespace.

### DIFF
--- a/lib/annotate.rb
+++ b/lib/annotate.rb
@@ -17,8 +17,6 @@ rescue StandardError
 end
 
 module Annotate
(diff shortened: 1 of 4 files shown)


KelseyDH:
Nice refactor that appears to limit the availability of `TRUE_RE` from bleeding into other ruby code. 👍

drwl:
@KelseyDH thanks! I appreciate the compliment.

PR #654, opened 2019-09-26
Make --additional_file_patterns consistent by using dashes

drwl:
Changed references to `--additional_file_patterns` to be `--additional-file-patterns` to make it consistent with other flags.


drwl:
@ryanwjackson tagging you just for fyi

Issue #657, opened 2019-09-27
LoadError: cannot load such file -- annotate/parser


9mm:
2.7.x worked fine but 3.0 on rails 6 gives me an error:

```
LoadError: cannot load such file -- annotate/parser
  /Users/Z/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/annotate-3.0.0/bin/annotate:18:in `require'
  /Users/Z/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/annotate-3.0.0/bin/annotate:18:in `<top (required)>'
  /Users/Z/.rbenv/versions/2.6.4/bin/annotate:23:in `load'
  /Users/Z/.rbenv/versions/2.6.4/bin/annotate:23:in `<top (required)>'
```

## Commands

```
$ bundle exec annotate
```

## Version

- annotate version: 3.0
- rails version 6.0.0
- ruby version 2.6.4

kapso:
Yea getting the same issue with those versions.

coderberry:
Same

TedTran2019:
Oh man, I thought I was insane. I was trying everything to get it to work, but it wouldn't.

I'm on Ruby 2.6.4, Rails 6.0, and annotate 3.0.0 and I get the exact same error.

drwl:
Will take a look, thanks for the report.

drwl:
Will close this when `3.0.1` gets released.

drwl:
Released `3.0.2`, @9mm can you confirm?

9mm:
I’m not at computer right now but I tested yesterday using master branch and it worked :) so I assume it’s good now. Thanks!

9mm:
It works 👌🏼

PR #662, opened 2019-09-29
Use a less error-prone way of specifying gem files

mattbrictson:
Before, whenever a new file is added to `lib`, the contributor or maintainer has to remember to explicitly add it to the gemspec. If this step is forgotten, it results in a broken release, like #657.

This commit makes this process less error prone by using `git ls-files` to automatically include all appropriate files in the gem. This is in line with what is advocated by Bundler's gem template:

https://github.com/bundler/bundler/blob/2a4d88a7044528209321fd4441e995ec66f604b9/lib/bundler/templates/newgem/newgem.gemspec.tt#L26

To test this PR I ran `rake gem` and verified that the list of files included in the gem is the same as before:

```
AUTHORS.rdoc
CHANGELOG.rdoc
README.rdoc
RELEASE.rdoc
TODO.rdoc
annotate.gemspec
bin/annotate
lib/annotate.rb
lib/annotate/active_record_patch.rb
lib/annotate/annotate_models.rb
lib/annotate/annotate_routes.rb
lib/annotate/constants.rb
lib/annotate/parser.rb
lib/annotate/tasks.rb
lib/annotate/version.rb
lib/generators/annotate/USAGE
lib/generators/annotate/install_generator.rb
lib/generators/annotate/templates/auto_annotate_models.rake
lib/tasks/annotate_models.rake
lib/tasks/annotate_models_migrate.rake
lib/tasks/annotate_routes.rake
```

### DIFF
--- a/annotate.gemspec
+++ b/annotate.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.email = ['alex@stinky.com', 'cuong.tran@gmail.com', 'x@nofxx.com', 'turadg@aleahmad.net', 'jon@cloudability.com']
   s.executables = ['annotate']
   s.extra_rdoc_files = ['README.rdoc', 'CHANGELOG.rdoc', 'TODO.rdoc']
-  s.files = `git ls-files -z *.rdoc *.gemspec bin lib`.split("\x0")
+  s.files = `git ls-files -z LICENSE.txt *.rdoc *.gemspec bin lib`.split("\x0")
   s.homepage = 'http://github.com/ctran/annotate_models'
   s.licenses = ['Ruby']
   s.require_paths = ['lib']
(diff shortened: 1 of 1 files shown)


drwl:
@mattbrictson oh wow this is neat, thanks for the PR. Having this would have definitely prevented an issue like #657. Two things:

1) looks like it's just missing `LICENSE.txt`, should be an easy fix.
2) what's the difference between using the `-z` flag and splitting on `\x0` compared to not using `-z` flag and splitting on `\n`, for example like how [Pry](https://github.com/pry/pry/blob/master/pry.gemspec#L35) does it?

mattbrictson:
> looks like it's just missing LICENSE.txt, should be an easy fix.

Nice catch, I'll add it.

> what's the difference between using the -z flag and splitting on \x0 compared to not using -z flag and splitting on \n, for example like how Pry does it?

I assume that if you run the code on a Windows machine then the line ending will not be `\n`, so it would break in that environment. The `-z` is therefore more robust across platforms.

Alternatively you could `require "English"` and then split on `$RS` (the platform-specific line separator), but the `-z` trick is a lot more concise.

drwl:
Gotcha thanks for the explanation, I learned something :).

Issue #663, opened 2019-09-30
trouble with annotate after upgrading from 2.7.5 to 3.0.2


gingerlime:
I'm trying to upgrade from 2.7.5 to 3.0.2 but somehow it looks like annotate fails silently and doesn't work any more...

A couple of things:

* our CI pipeline runs `bundle exec annotate` to make sure no model annotations are missing. This now fails. Perhaps annotate no longer outputs "Model files unchanged" when there are no changes? I'm not sure
* I tried to create a simple migration to a table and running `rake db:migrate`, but it doesn't automatically add annotations (no errors, I did add the rake task using `rails g annotate:install`).
* also manually running `bundle exec annotate` after the migration was run (I can see `schema.rb` updated), doesn't update the annotations... It just doesn't seem to do anything... No errors or exit code other than 0 either.

## Commands

```
$ git st
On branch dependabot/bundler/annotate-3.0.2
Your branch is up to date with 'origin/dependabot/bundler/annotate-3.0.2'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        new file:   db/migrate/20190930085602_blablabla.rb
        new file:   lib/tasks/auto_annotate_models.rake

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   db/schema.rb

$ git diff db/schema.rb
diff --git a/db/schema.rb b/db/schema.rb
index 4ddfcf5e8..3d1173ce6 100644
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.

-ActiveRecord::Schema.define(version: 2019_08_22_094809) do
+ActiveRecord::Schema.define(version: 2019_09_30_085602) do

   # These are extensions that must be enabled in order to support this database
   enable_extension "hstore"
@@ -469,6 +469,7 @@ ActiveRecord::Schema.define(version: 2019_08_22_094809) do
     t.string "active_session_id"
+    t.boolean "xyz"

...

$ bundle exec annotate
warning: parser/current is loading parser/ruby25, which recognizes
warning: 2.5.6-compliant syntax, but you are running 2.5.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
[...]

PR #671, opened 2019-10-14
Defaults annotate models to true

rhiroshi:
Setting the default to annotate models as it is the main feature of the gem.
Solves #663


drwl:
Thanks for the PR!

PR #672, opened 2019-10-15
Get rid of vertical alignment in annotation options generator

JanaVPetrova:
Hello! Recently we updated Annotate in our project and we had to generate the configuration file again. The diff was pretty large and it was hard to understand what options changed because of the vertical alignment. I know that there are different opinions on vertical alignment, but I think it would be better without it here: 
- people will not have to maintain it 
- when upgrading the library the diff in configuration will be more readable


ctran:
Thanks but I rather not change this at this point since it would cause more conflicts for current folks.  By the way, you should be able to tell diff to ignore white space changes.

PR #678, opened 2019-10-26
[Fix #570] Change of foreign key should be considered as a column change

### LINKED ISSUE #570: foreign key change of type don't get noticed
## Commands

When I change foreign key type from :restrict to :cascade or back
in e.g. Rails migration:
```
      remove_foreign_key :imports, :clients
      add_foreign_key :imports, :clients, on_delete: :cascade
```
then run
`bundle exec annotate`
I get 
`Model files unchanged.`
And it's at outdated
```
# Foreign Keys
#
#  fk_rails_...  (client_id => clients.id) ON DELETE => restrict
```
Only running with --force updates them properly.

## Version

- annotate v2.7.4
- Rails 5.2.0
- ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]

sashabelozerov:
Fixes #570

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -515,7 +515,7 @@ module AnnotateModels
       old_header = old_content.match(header_pattern).to_s
       new_header = info_block.match(header_pattern).to_s
 
-      column_pattern = /^#[\t ]+[\w\*`]+[\t ]+.+$/
+      column_pattern = /^#[\t ]+[\w\*\.`]+[\t ]+.+$/
       old_columns = old_header && old_header.scan(column_pattern).sort
       new_columns = new_header && new_header.scan(column_pattern).sort
 
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -414,7 +414,7 @@ Metrics/AbcSize:
 # Configuration parameters: CountComments, ExcludedMethods.
 # ExcludedMethods: refine
 Metrics/BlockLength:
-  Max: 244
+  Max: 259
 
 # Offense count: 1
 # Configuration parameters: CountBlocks.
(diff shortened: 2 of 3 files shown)


lib/annotate/annotate_models.rb
```diff
@@ -515,7 +515,7 @@ def annotate_one_file(file_name, info_block, position, options = {})
       old_header = old_content.match(header_pattern).to_s
       new_header = info_block.match(header_pattern).to_s
 
-      column_pattern = /^#[\t ]+[\w\*`]+[\t ]+.+$/
+      column_pattern = /^#[\t ]+[\w\*\.`]+[\t ]+.+$/
```
ctran: Could you leave a note as to why this would fix the problem?

sashabelozerov: @ctran Line with a foreign key info in an annotation text looks like:

#  fk_rails_...  (foreign_thing_id => foreign_things.id) ON DELETE => restrict

It didn't match the previous pattern because of dots at the end of the foreign key name (we have dots there because we truncate generated foreign key names at https://github.com/ctran/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L477).

In turn, because a foreign key line didn't match the pattern, we didn't check it for changes at https://github.com/ctran/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L525

Hope this helps.

ctran: Does it mean if the name doesn't end with "...", this pattern won't match anymore?

sashabelozerov: @ctran No. ```[\w\*\.`]``` - this part of the regexp represents characters which a name can consist of, in any order. It will match both names with "..." and without them.

ctran:
Thanks!

PR #682, opened 2019-11-09
Revert behavior of Github release action

drwl:
Reverted tag matcher to to use `v*` prefix instead of `release/v*`.


drwl:
Just fyi @ctran

ctran:
Thanks, I did that for testing purpose only.

drwl:
👍 I'm planning on migrating off of Travis so we can just be fully Github. There's an issue and request for help in #683.

PR #692, opened 2019-11-21
ActiveAdmin uses Pluralized file names

danielricecodes:
Changing ActiveAdmin code so it reads pluralized model named files.  This is how active admin generates the files when using its own generator, so I think this gem ought to follow what the gem produces in its own generator.

Ex:

`rails g active_admin:resource Foo` will create a file with a pluralized file name `app/admin/foos.rb`.


drwl:
Thanks @danielricecodes for the PR and good description. Would it be possible to add a test for this?

drwl:
Closing in favor of #776

PR #693, opened 2019-11-21
Add ActiveAdmin option to template

danielricecodes:
Add's the `active_admin` option to the auto annotate rake task template.


drwl:
@danielricecodes thanks for the patch! I'll be merging it in.

PR #695, opened 2019-11-23
Fix foreign key issue with Rails 6 and Sqlite3

### LINKED ISSUE #620: undefined method `gsub' for nil:NilClass
Hi!

First of all, thank you for this really useful gem that is a must in my Rails projects.

I've been messing up with Rails 6, and when I created a new migration/model and tried to annotate the files, it seems something got wrong, as you can see in the error messages below.

Here's my migration file:

```ruby
class CreateSheep < ActiveRecord::Migration[6.0]
  def change
    create_table :sheep do |t|
      t.string :name
      t.date :date_of_birth
      t.references :address, null: true, foreign_key: true
      t.string :phone_number

      t.timestamps
    end
  end
end

```

And the model:

```ruby
class Sheep < ApplicationRecord
  belongs_to :address
end
```

The migration ran fine, but is it something wrong I'm doing?

## Commands

```sh
$ bundle exec annotate
#> Unable to annotate app/models/sheep.rb: undefined method `gsub' for nil:NilClass
#> Unable to annotate app/models/sheep.rb: no implicit conversion of nil into Array
#> Model files unchanged.
```

## Version

- annotate version: tried with branch develop and master
- rails version: 6.0.0.rc1
- ruby version: 2.6.3

shime:
Rails 6 has dropped name from options for ActiveRecord::ConnectionAdapters::ForeignKeyDefinition
for Sqlite3 adapter:

https://github.com/rails/rails/blob/f2df77709f7e536aaf4d6f984ff21a49d44d34c1/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L305-L311

This has caused issues with getting foreign key info and blowing up
when trying to annotate models with foreign keys. This commit
adds one additional check for name presence and uses column
attribute from options instead.

Closes #620.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -472,7 +472,10 @@ module AnnotateModels
       foreign_keys = klass.connection.foreign_keys(klass.table_name)
       return '' if foreign_keys.empty?
 
-      format_name = ->(fk) { options[:show_complete_foreign_keys] ? fk.name : fk.name.gsub(/(?<=^fk_rails_)[0-9a-f]{10}$/, '...') }
+      format_name = lambda do |fk|
+        return fk.options[:column] if fk.name.blank?
+        options[:show_complete_foreign_keys] ? fk.name : fk.name.gsub(/(?<=^fk_rails_)[0-9a-f]{10}$/, '...')
+      end
 
       max_size = foreign_keys.map(&format_name).map(&:size).max + 1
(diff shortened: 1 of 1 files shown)


drwl:
@shime thanks for the patch. It looks good to me. Just for my understanding, can you elaborate more on the code you linked to? 

v5.2 looks to have the same code 
https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L356

PR #698, opened 2019-12-16
Upgrade and fix CI

drwl:
Looks like it's breaking in Github CI because Ruby 2.3.X is no longer supported by `actions/setup-ruby@v1`. Ruby 2.3.X has stopped being supported since March of 2019, so I've removed Ruby 2.3.X versions from both Github and Travis CI.

It's also breaking in Travis CI because it's timing out in an interactive prompt when trying to do `gem update --system`. This is due to changes to [Ruby Gems 3.1.1](https://github.com/rubygems/rubygems/issues/3036). This has been fixed.

Failing Github CI build: https://github.com/ctran/annotate_models/runs/335456181
Failing Travis build: https://api.travis-ci.org/v3/job/625759725/log.txt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - 2.3.x
           - 2.4.x
           - 2.5.x
           - 2.6.x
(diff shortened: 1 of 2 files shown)


drwl:
Looks like there's an open issue on RubyGems github project https://github.com/rubygems/rubygems/issues/3036

PR #707, opened 2019-12-25
Refactor lib/annotate.rb

drwl:
`lib/annotate.rb` is bloated. This PR refactors methods out of it into `Annotate::Helpers` and adds test coverage for the methods.


lib/tasks/annotate_models.rake
```diff
@@ -11,46 +11,46 @@ task annotate_models: :environment do
   require "#{annotate_lib}/annotate/active_record_patch"
 
   options = {is_rake: true}
-  ENV['position'] = options[:position] = Annotate.fallback(ENV['position'], 'before')
+  ENV['position'] = options[:position] = Annotate::Helpers.fallback(ENV['position'], 'before')
   options[:additional_file_patterns] = ENV['additional_file_patterns'] ? ENV['additional_file_patterns'].split(',') : []
```
ctran: Can this be mixed in instead of specifying full path?  It would also cut back the number of changes needed.

drwl: @ctran this is done intentionally for now to make it easier to see where everything is coming from. Do you think it would be okay to keep for now? 

I'm thinking we could untangle `lib/annotate.rb`. I'm looking at the Pry gem as a source of inspiration.

ctran: I'm fine either way, was just wondering if you had specific reasons to do that.

drwl: Just to make it more explicit where it's coming from right now. I'm expecting `Helpers` to be a temporary solution until it's clearer what the different pieces are.

PR #711, opened 2020-01-07
Add constants MAGIC_COMMENT_MATCHER

nard-tech:
I added `AnnotateModels::MAGIC_COMMENT_MATCHER` and `AnnotateRoutes::MAGIC_COMMENT_MATCHER` instad of `AnnotateModels.magic_comment_matcher` and `AnnotateRoutes.magic_comment_matcher`.

I think that they should be constants because the result of Regexp is always fixed.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -82,6 +82,8 @@ module AnnotateModels
     }
   }.freeze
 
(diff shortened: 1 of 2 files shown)


nard-tech:
The Regexp in AnnotateModels and that in AnnotateRoutes are different. Is it OK?

drwl:
> The Regexp in AnnotateModels and that in AnnotateRoutes are different. Is it OK?
> 
> I think that `AnnotateModels::MAGIC_COMMENT_MATCHER` and `AnnotateRoutes::MAGIC_COMMENT_MATCHER` may be unified to another constant (e.g. `Annotate::MAGIC_COMMENT_MATCHER`).

@nard-tech It should be okay that they're different for the time being. I think without better test coverage we shouldn't make behavior changes.

nard-tech:
@drwl 
> I think without better test coverage we shouldn't make behavior changes.

All right. Thank you for your reply.

PR #713, opened 2020-01-08
Freeze constant AnnotateRoutes::HEADER_ROW

nard-tech:
As the change log shows, I froze the constant `HEADER_ROW` in accordance with the normal Ruby style regulation.


drwl:
Maybe in a follow up step could we use the magic string? `# frozen_string_literal: true`

nard-tech:
@drwl Oh, exactly it is better to add the magic comment `# frozen_string_literal: true`.
I'll try it in another PR.

PR #714, opened 2020-01-08
Refactor AnnotateRoutes.header

nard-tech:
I refactored `AnnotateRoutes.header` for performance, and add `AnnotateRoutes.comment` to make code DRY.

### DIFF
--- a/lib/annotate/annotate_routes.rb
+++ b/lib/annotate/annotate_routes.rb
@@ -86,10 +86,10 @@ module AnnotateRoutes
       end
       out << '' if magic_comments_map.any?
 
-      out += ["# #{options[:wrapper_open]}"] if options[:wrapper_open]
+      out << "# #{options[:wrapper_open]}" if options[:wrapper_open]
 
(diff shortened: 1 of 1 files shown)


lib/annotate/annotate_routes.rb
```diff
@@ -86,29 +86,37 @@ def header(options = {})
       end
       out << '' if magic_comments_map.any?
 
-      out += ["# #{options[:wrapper_open]}"] if options[:wrapper_open]
+      out << comment(options[:wrapper_open]) if options[:wrapper_open]
 
-      out += ["# #{options[:format_markdown] ? PREFIX_MD : PREFIX}" + (options[:timestamp] ? " (Updated #{Time.now.strftime('%Y-%m-%d %H:%M')})" : '')]
-      out += ['#']
+      out << comment(options[:format_markdown] ? PREFIX_MD : PREFIX) + (options[:timestamp] ? " (Updated #{Time.now.strftime('%Y-%m-%d %H:%M')})" : '')
```
drwl: In looking at this line, it seems kinda strange and hard to read.

I'm sure there's a lot of this going on in the code base.

nard-tech: @drwl I agree with you. I want to try to fix it, but I think that it is better to add test cases before refactor this more.

PR #719, opened 2020-01-13
Refactor namespace Annotate

nard-tech:
#   Summary

*   Remove redundant `Annotate.loaded_tasks=` and `Annotate.loaded_tasks`
*   Remove unnecessary comment for Rubocop
*   Make Annotate.load_requires private
*   Refactor `Annotate.load_tasks`

### DIFF
--- a/lib/annotate.rb
+++ b/lib/annotate.rb
@@ -72,17 +72,9 @@ module Annotate
     options
   end
 
-  def self.loaded_tasks=(val)
-    @loaded_tasks = val
-  end
-
-  def self.loaded_tasks
-    @loaded_tasks
-  end
-
   def self.load_tasks
-    return if loaded_tasks
-    self.loaded_tasks = true
+    return if @loaded_tasks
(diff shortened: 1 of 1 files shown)


lib/annotate.rb
```diff
@@ -136,4 +131,13 @@ def self.bootstrap_rake
     load_tasks
     Rake::Task[:set_annotation_options].invoke
   end
+
+  class << self
```
drwl: This seems interesting, I'm not sure I've seen this before. Could you explain what is supposed to be happening? 

Seems like a module that has class-level methods. But I'm confused as to what happens when doing the
```ruby
class << self
...
end
```

in a module

nard-tech: @drwl when you want to defined private method, you can write like this.

https://stackoverflow.com/questions/4952980/how-to-create-a-private-class-method/12311586#12311586

drwl: Gotcha, thanks for the link

PR #721, opened 2020-01-16
Rename FactoryGirl to FactoryBot

nard-tech:
Now the gem [factory_girl](https://github.com/thoughtbot/factory_girl) have been renamed to [factory_bot](https://github.com/thoughtbot/factory_bot).

So I fixed constant names and comments.


contradictioned:
Seems like this did not fix everything? See https://github.com/ctran/annotate_models/blob/develop/lib/annotate/annotate_models.rb#L145 for example.

PR #723, opened 2020-01-17
Refactor AnnotateModels::Helpers

nard-tech:
I refactored two methods in AnnotateModels::Helpers for readability.

### DIFF
--- a/lib/annotate/helpers.rb
+++ b/lib/annotate/helpers.rb
@@ -19,7 +19,7 @@ module Annotate
       end
 
       def fallback(*args)
-        args.detect { |arg| !arg.blank? }
+        args.detect(&:present?)
       end
 
       def reset_options(options)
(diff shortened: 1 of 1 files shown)


lib/annotate/helpers.rb
```diff
@@ -15,10 +15,7 @@ def include_models?
       end
 
       def true?(val)
-        return false if val.blank?
-        return false unless val =~ Constants::TRUE_RE
-
-        true
+        val.present? && Constants::TRUE_RE.match?(val)
```
nard-tech:
```ruby
def true?(val)
  return false if val.blank?
  return false unless val =~ Constants::TRUE_RE
  true
```
can be refactored to
```ruby
def true?(val)
  if val.blank?
    false
  else
    if val =~ Constants::TRUE_RE
      true
    else
      false
    end
  end
end
```
So we can easily replace them to a one liner method.

I used `Regexp#match?` instead of `String#=~` because `String#=~` returns `nil` or an integer.

drwl: 👍 nice

PR #724, opened 2020-01-18
Support YARD notation

tvallois:
Hi,

The scope of this pull request is to allow annotate_models to generate models documentation using YARD. This is the first step, i'll add more features later.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -304,6 +304,14 @@ module AnnotateModels
                    end
         if options[:format_rdoc]
           info << sprintf("# %-#{max_size}.#{max_size}s<tt>%s</tt>", "*#{col_name}*::", attrs.unshift(col_type).join(", ")).rstrip + "\n"
+        elsif options[:format_yard]
+          info << sprintf("# @!attribute #{col_name}") + "\n"
+          if col_type == 'boolean'
+            info << sprintf("#   @return [TrueClass]") + "\n"
+            info << sprintf("#   @return [FalseClass]") + "\n"
+          else
+            info << sprintf("#   @return [#{map_col_type_to_ruby_classes(col_type)}]") + "\n"
+          end
         elsif options[:format_markdown]
           name_remainder = max_size - col_name.length - non_ascii_length(col_name)
           type_remainder = (md_type_allowance - 2) - col_type.length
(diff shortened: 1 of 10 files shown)


lib/annotate/annotate_models.rb
```diff
@@ -904,6 +904,17 @@ def classified_sort(cols)
       ([id] << rest_cols << timestamps << associations).flatten.compact
     end
 
+    def map_col_type_to_ruby_classes(col_type)
```
drwl: Just curious - can you elaborate why it should not be private and public instead?

lib/annotate/annotate_models.rb
```diff
@@ -957,6 +957,7 @@ def map_col_type_to_ruby_classes(col_type)
     when 'datetime', 'timestamp', 'time'                 then Time.to_s
     when 'date'                                          then Date.to_s
     when 'text', 'string', 'binary', 'inet', 'uuid'      then String.to_s
+    when 'json', 'jsonb'                                 then Hash.to_s
```
drwl: Can you elaborate on this?

lib/annotate/annotate_models.rb
```diff
@@ -913,6 +908,7 @@ def map_col_type_to_ruby_classes(col_type)
       when 'date'                                          then Date.to_s
       when 'text', 'string', 'binary', 'inet', 'uuid'      then String.to_s
       when 'json', 'jsonb'                                 then Hash.to_s
+      when 'boolean'                                       then 'Boolean'
```
drwl: Does this change any current behavior?

lib/annotate/annotate_models.rb
tvallois: I don't know, my first idea was that it can be reused somewhere else but seems not yet so i'll add it in private

lib/annotate/annotate_models.rb
tvallois: Seems not. Boolean is a class that YARD understand quite well. It's better than TrueClass and FalseClass

Samsinite: If you are referring to just the boolean case, `Boolean` is the string for what YARD expects a boolean type to be in the comments.

lib/annotate/annotate_models.rb
tvallois: When a model has a json/jsonb column, it's transformed in Hash class in ActiveRecord. That's the same logic i've followed here.

lib/annotate/annotate_models.rb
drwl: Gotcha

drwl:
@tvallois looks like it's good to go. If you can resolve merge conflicts I'm happy to merge this in next 🙂

tvallois:
> @tvallois looks like it's good to go. If you can resolve merge conflicts I'm happy to merge this in next

done :)

Samsinite:
🎉 excited for this!

PR #725, opened 2020-01-19
Remove unworking integration tests

drwl:
`spec/integration` contains a lot of files that were used for integration tests. They stopped were [disabled in 2014](https://github.com/ctran/annotate_models/commit/9540121243da58cad3ab3c0ba444f22445a995ed#diff-78ddf877ecc2a9344997ef077a77955a) and haven't been working since.

Removing because those tests don't run, don't work, and are outdated. The plan is to re-introduce integration tests sometime in the future.


drwl:
@ctran any thoughts against merging this? I'll add an issue for help-wanted to add integration tests to make sure we keep track.

drwl:
@ctran gonna merge this in and if we decide we want to keep it we can revert this commit.

PR #718, opened 2020-01-12
Refactor RSpec for AnnotateModels

nard-tech:
I refactored and structuralized RSpec test cases of AnnotateModels for readability and scalability because it was too complex to read.

PR #726, opened 2020-01-19
Refactor RSpec for AnnotateModels (1)

nard-tech:
I refactored and structuralized RSpec test cases of AnnotateModels for readability and scalability because it was too complex to read.

cf. #718 
In this PR, I refactored test cases of some methods in `AnnotateModels`.
I will refactor test cases of other methods in another PR.

### DIFF
--- a/spec/lib/annotate/annotate_models_spec.rb
+++ b/spec/lib/annotate/annotate_models_spec.rb
@@ -2305,12 +2305,14 @@ EOS
       allow(Foo).to receive(:table_exists?) { false }
     end
 
+    subject do
+      AnnotateModels.annotate_model_file([], 'foo.rb', nil, {})
+    end
+
     after { Object.send :remove_const, 'Foo' }
 
     it 'skips attempt to annotate if no table exists for model' do
-      annotate_model_file = AnnotateModels.annotate_model_file([], 'foo.rb', nil, {})
-
-      expect(annotate_model_file).to eq nil
+      is_expected.to eq nil
     end
   end
 end
--- a/spec/lib/annotate/annotate_models_spec.rb
+++ b/spec/lib/annotate/annotate_models_spec.rb
@@ -1539,187 +1539,236 @@ EOS
     end
   end
 
-  describe '#remove_annotation_of_file' do
-    require 'tmpdir'
-
-    def create(file, body = 'hi')
-      path = File.join(@dir, file)
-      File.open(path, 'w') do |f|
-        f.puts(body)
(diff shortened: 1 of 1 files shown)


spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -1831,9 +1831,9 @@ class Foo < ActiveRecord::Base
   describe 'annotating a file' do
     before do
       @model_dir = Dir.mktmpdir('annotate_models')
-      (@model_file_name, @file_content) = write_model 'user.rb', <<-EOS
-class User < ActiveRecord::Base
-end
+      (@model_file_name, @file_content) = write_model 'user.rb', <<~EOS
```
drwl: I looked up what the differences are in the doc string `<<-` vs `<<~`. Learning something new 👍

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -925,16 +925,16 @@ def mock_column(name, type, options = {})
...
 
-        it do
-          expect(subject).to eq(additional_file_patterns)
+        it 'returns additional_file_patterns in the argument "options"' do
+          is_expected.to eq(additional_file_patterns)
         end
       end
 
-      context 'without additional_file_patterns' do
+      context 'when additional_file_patterns is not specified is the options' do
```
drwl: Some typos here, but we can clean that up in a later commit :)

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -894,16 +894,16 @@ def mock_column(name, type, options = {})
...
```
drwl: Another typo here

spec/lib/annotate/annotate_models_spec.rb
nard-tech: Thanks, I fixed it in #731.

spec/lib/annotate/annotate_models_spec.rb
nard-tech: cf. https://www.rubyguides.com/2018/11/ruby-heredoc/

I like `<<~` because it makes code more clean.
When using VS Code, both strings indented with `<<~` and that of `<<-` can be collapsed, but the latter looks bad.

drwl: Yep the `<<~` does look more clean. I think that it may have not existed when that spec was written but now it does 😊

PR #735, opened 2020-01-20
Refactor RSpec for AnnotateModels (4) - AnnotateModels.get_schema_info (without custom options)

nard-tech:
I refactored test cases of AnnotateModels without custom options.

I will refactor the test case added in #602 in another PR.


drwl:
@nard-tech what does changing the position of test cases do?

nard-tech:
@drwl After #741, I'd like to separate `.get_schema_info` to another namespace because `AnnotateModels` is too fat to maintain.
At that time I will change the position of test cases.

PR #739, opened 2020-01-22
Fix constant names that were not renamed in #721

contradictioned:
As commented in #721, it seems like not all occurrences of the constants were renamed. This proposes a fix.


nard-tech:
@contradictioned Thank you. I carelessly dropped some commit when I rebased branch of #721.

drwl:
Thanks for the PR @contradictioned and catching this

PR #741, opened 2020-01-27
Refactor RSpec for AnnotateModels (5) - structuralize test cases

nard-tech:
The test cases of `AnnotateModels.get_schema_info` was separated into two clauses.
So I merged them into one clause, and refactor test cases.

### DIFF
--- a/spec/lib/annotate/annotate_models_spec.rb
+++ b/spec/lib/annotate/annotate_models_spec.rb
@@ -176,21 +176,21 @@ describe AnnotateModels do
   end
 
   describe '.get_schema_info' do
+    let :klass do
+      mock_class(:users, primary_key, columns, indexes, foreign_keys)
+    end
+
+    let :indexes do
+      []
+    end
+
+    let :foreign_keys do
+      []
+    end
+
     context 'when option is not present' do
       subject do
         AnnotateModels.get_schema_info(klass, header)
-      end
-
-      let :klass do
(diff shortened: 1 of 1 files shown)


spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -176,128 +176,38 @@ def mock_column(name, type, options = {})
...
-
-    let :foreign_keys do
-      []
-    end
-
-    context 'when header is "Schema Info"' do
-      let :header do
-        'Schema Info'
+    context 'when option is not present' do
+      subject do
```
drwl: Any reason why is `subject` under `context` instead of `describe`?

nard-tech: @drwl Thanks, I'd like to split large commits into smaller ones.

I use `context` and `describe` depending on following link.
https://lmws.net/describe-vs-context-in-rspec

Are you getting at that `subject` under `context` is strange and `subject` should be under `describe`?

drwl: Yep, I've usually seen `subject` be under `describe` instead of `context` but there doesn't seem to be anything on the style guide against it.

https://github.com/rubocop-hq/rspec-style-guide#subject-naming-in-context

Has `subject` under multiple contexts so seems to be fine.

nard-tech: @drwl All right. I'll use your reply as a reference in the next PR. Thank you.

PR #747, opened 2020-02-01
Add integration tests to project

### LINKED ISSUE #733: Add integration framework to project
## Summary
This project can be used in many different environments and scenarios. Can be run using different versions of Ruby across different versions of Rails. Having integration tests that can help this project move forward and stay up to date when new versions come up.

Would like to add or fix these type of tests. Currently there exist [non-working integration tests](https://github.com/ctran/annotate_models/pull/725) that haven't been functional since ~2014. A lot has changed since then. Bundler has come a long way for managing dependencies on projects and CI has improved a lot so we may be able to [leverage matrix builds](https://docs.travis-ci.com/user/build-matrix/).

## Next Steps
Having a working bare bones integration test (this will take a good amount of effort)

drwl:
Adds integration tests to the project that's able to run in CI across different ruby versions.

* Adds a Rails 5.2.4.1 app
* Adds a Rails 6.0.2.1 app

Fixes #733


drwl:
@ctran added if you wanna take a look

drwl:
@nard-tech could you take a look and tell me what you think?

nard-tech:
It looks good and there is no problem.

Will you add tests for `AnnotateRoutes`?

drwl:
> @drwl
> It looks good and there is no problem.
> Will you add tests for `AnnotateRoutes`?

Sure will do

drwl:
Note for any future visitors, ad4f9bad7836cec6dfe238e73d1ab8c1c01f7e62 does something hacky which is to have the constants be uniquely named. The constants defined in the `describe` block become part of the global namespace. 

`let` statements cannot be used because rspec complains about reading values from `let` statements in a `before(:all)` block.

Definitely an area to improve.

PR #748, opened 2020-02-02
Refactor test cases for AnnotateRoutes.remove_annotations

nard-tech:
Refactor of test cases for AnnotateRoutes.remove_annotations after #736.

### DIFF
--- a/spec/lib/annotate/annotate_routes_spec.rb
+++ b/spec/lib/annotate/annotate_routes_spec.rb
@@ -319,96 +319,116 @@ describe AnnotateRoutes do
     end
   end
 
-  describe 'When removing' do
-    before(:each) do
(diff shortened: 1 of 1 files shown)


spec/lib/annotate/annotate_routes_spec.rb
```diff
@@ -319,96 +319,116 @@
...
-        # My comment
-        # == Route Map
-        #
-        # another good line
-        # good line
-        Rails.application.routes.draw do
-          root 'root#index'
-        end
-      EOS
+    context 'When custome comments are above route map' do
```
drwl: typo `custome`?

nard-tech: Thank you for your pointing out my mistake.
I fixed it.

PR #749, opened 2020-02-02
Refactor RSpec for AnnotateModels - with Globalize gem

nard-tech:
Refactor test cases for AnnotateModels.get_schema_info with Globalize gem


nard-tech:
I will split #741 into several PRs.

Issue #674, opened 2019-10-17
Wrong table annotations for models with same name, but different tables


jimmytang:
Describe your problem here.
Two different models, both with the same class name. One is inside a Rails engine inside the app, the other inside the main rails app. They are pointing to different underlying PostGres tables. 

When running model annotations, both models show the annotations for the model inside the main app. 

The expected result should be the model inside the main app has annotations representing it's underlying table and the model inside the engine has a different set of annotations representing it's underlying table.

## Commands
```
rake db:update_model_annotations
```

## Version
- annotate version
2.7.5
- rails version
5.2.3
- ruby version
2.6.2p47

maxh:
Looks like a fix for this might be adding an explicit self.table_name = "foo_bar" to one of the models.

MGPalmer:
> Looks like a fix for this might be adding an explicit `self.table_name = "foo_bar"` to one of the models.

We've got the same issue, and this doesn't work for us :(

PR #767, opened 2020-02-27
Adding option to skip loading models from subdirectories

ethanbresler:
Currently the models annotator automatically attempts to find a class with a matching name at the bottom of project's directory tree before going up into specific engine's models.  This causes issues with models that share names with classes in other engines or lower classes in the project's directory.   This PR adds the option to skip attempts to load classes from lower directories and just uses the model's file path directly. 

 #674 

cc @maxh

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -631,7 +631,7 @@ module AnnotateModels
     def parse_options(options = {})
       self.model_dir = split_model_dir(options[:model_dir]) if options[:model_dir]
       self.root_dir = options[:root_dir] if options[:root_dir]
-      self.skip_subdirectory_model_load = options[:skip_subdirectory_model_load].present? ? options[:skip_subdirectory_model_load] : false
+      self.skip_subdirectory_model_load = options[:skip_subdirectory_model_load].present?
     end
 
     def split_model_dir(option_value)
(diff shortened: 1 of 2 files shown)


lib/annotate/annotate_models.rb
```diff
@@ -769,6 +784,7 @@ def get_loaded_model_by_path(model_path)
     def parse_options(options = {})
       self.model_dir = split_model_dir(options[:model_dir]) if options[:model_dir]
       self.root_dir = options[:root_dir] if options[:root_dir]
+      self.skip_subdirectory_model_load = options[:skip_subdirectory_model_load].present? ? options[:skip_subdirectory_model_load] : false
```
nard-tech:
```suggestion
self.skip_subdirectory_model_load = options[:skip_subdirectory_model_load].present?
```
It seems to be better, I think.

ethanbresler: Awesome, didn't realize you could do that with `present?`

drwl:
@ethanbresler I apologize for the delay in reviewing. Thanks for this patch. Would it be possible to add tests for this?

ethanbresler:
> @ethanbresler I apologize for the delay in reviewing. Thanks for this patch. Would it be possible to add tests for this?

@drwl Added tests, but the CI tests are failing due to an 'install sql lite' error.  Is this something on my end or a problem with the integration tests?

maxh:
@drwl can we get another review of this? thanks!

drwl:
@maxh I apologize for the delay. @ethanbresler can you try rebasing? I believe that should be fixed.

ethanbresler:
@drwl rebased and checks are passing.  Thanks for your help with this.


cc @maxh

ethanbresler:
@drwl any updates here?  Can I get this merged?  Thanks so much!

PR #776, opened 2020-03-12
ActiveAdmin: Allow for both singular and plural model names

vfonic:
This is an addition to this PR: https://github.com/ctran/annotate_models/pull/692
I've seen it's been stale for a while and it's rather simple fix.

I noticed my ActiveAdmin resources are not being annotated.
I know ActiveAdmin uses plural model names, but, as far as I remember, they used to use singular model names in the past.

We can see some "evidence" of that here: https://activeadmin.info/2-resource-customization.html#customizing-parent-menu-items
Probably in other places as well.

In order to keep the gem working as expected for singular ActiveAdmin resources, I suggest we annotate both singular and plural model names.

I just tested this out in a project of mine where I renamed one file to be in singular form and left the other files in plural form.
It worked like a charm. :)

I also added a test, as best to my knowledge, as there are no similar tests that I could find.

Please let me know what you think!

/cc @danielricecodes @drwl 

Also please, don't make the default rake task do a git checkout! I've never seen this and I've lost the above changes once due to this! I just didn't expect that running tests would be changing the source files in any way.

### DIFF
--- a/lib/annotate/annotate_models/file_patterns.rb
+++ b/lib/annotate/annotate_models/file_patterns.rb
@@ -59,7 +59,10 @@ module AnnotateModels
         when 'controller'
           [File.join(root_directory, CONTROLLER_DIR, '%PLURALIZED_MODEL_NAME%_controller.rb')]
         when 'admin'
-          [File.join(root_directory, ACTIVEADMIN_DIR, '%MODEL_NAME%.rb')]
+          [
+            File.join(root_directory, ACTIVEADMIN_DIR, '%MODEL_NAME%.rb'),
+            File.join(root_directory, ACTIVEADMIN_DIR, '%PLURALIZED_MODEL_NAME%.rb')
+          ]
         when 'helper'
           [File.join(root_directory, HELPER_DIR, '%PLURALIZED_MODEL_NAME%_helper.rb')]
         else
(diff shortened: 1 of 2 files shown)


spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -1631,6 +1631,16 @@ def mock_column(name, type, options = {})
         end
       end
     end
+
+    context 'when pattern_type is "admin"' do
+      let(:root_directory) { '/root' }
+      let(:pattern_type) { 'admin' }
+      let(:options) { {} }
+
+      it 'returns an empty array' do
```
drwl, line 1640: Is this expectation description accurate?

vfonic: I think you're right. I've updated the description. Thanks!

vfonic:
I've updated the spec description, rebased on `develop` and force-pushed the amended commit.

vfonic:
I've resolved merge conflicts with the latest changes to `develop` again.

@drwl can you please have a look? Thanks!

vfonic:
@drwl I fixed the rubocop issues, ran rubocop locally to confirm, rebased over `develop`, force-pushed to my feature branch and now everything should be up to date. :)

drwl:
> Also please, don't make the default rake task do a git checkout! I've never seen this and I've lost the above changes once due to this! I just didn't expect that running tests would be changing the source files in any way.

@vfonic to clarify, did you mean that running the tests would clear your branch? If so, I apologize for that. It was a hacky approach to getting integration tests working. I've since made it opt-in via env variable. Once this project stabilizes a bit more I'll work on making it non-destructive.

vfonic:
@drwl Yes, that's what happened. Thanks for clarifying this! I'm glad to hear this wasn't a permanent solution. :)

PR #779, opened 2020-03-22
Fix output for multiline column comments

### LINKED ISSUE #778: Issue generating annotations when a column comment has newline character
When annotating a model with a column that had a comment that includes a newline character, the annotation will interpolate the newline character and add a non-commented line. 

For example:

```ruby
# schema.rb
  create_table "users", force: :cascade do |t|
    t.string "name", comment: "This is a comment.\nWith two lines!"
    t.datetime "created_at", precision: 6, null: false
    t.datetime "updated_at", precision: 6, null: false
  end
```
will generate the following annotation:

```ruby
# user.rb

# == Schema Information
#
# Table name: users
#
#  id                                       :bigint           not null, primary key
#  name(This is a comment.
With two lines!) :string
#  created_at                               :datetime         not null
#  updated_at                               :datetime         not null
#
```

## Commands

```
$ annotate --models --with-comment
```

## Version

- annotate version: 3.1.0
- rails version: 6.0.2.2
- ruby version: 2.6.5

tmr08c:
Closes #778 

If a column comment includes the newline character, the newline character
would be "printed" into the annotation block resulting in a line break
and an uncommented line.

For example, for the following table:

```
create_table "users", force: :cascade do |t|
  t.string "name", comment: "This is a comment.\nWith two lines!"
  t.datetime "created_at", precision: 6, null: false
  t.datetime "updated_at", precision: 6, null: false
end
```

annotating the model with the `--with-comment` flag will result in:

```
\# == Schema Information
\#
\# Table name: users
\#
\#  id                                       :bigint           not null, primary key
\#  name(This is a comment.
With two lines!) :string
\#  created_at                               :datetime         not null
\#  updated_at                               :datetime         not null
\#
```

This uncommented line would result in invalid Ruby and cause the file to
no longer be valid.

This fix replaces the newline character with an escaped version, so the
output will look more like:

```
\# == Schema Information
\#
\# Table name: users
\#
\#  id                                       :bigint           not null, primary key
\#  name(This is a comment.\nWith two lines!):string
\#  created_at                               :datetime         not null
\#  updated_at                               :datetime         not null
\#
```

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -251,10 +251,11 @@ module AnnotateModels
         col_type = get_col_type(col)
         attrs = get_attributes(col, col_type, klass, options)
         col_name = if with_comments?(klass, options) && col.comment
-                     "#{col.name}(#{col.comment})"
+                     "#{col.name}(#{col.comment.gsub(/\n/, "\\n")})"
                    else
                      col.name
                    end
+
         if options[:format_rdoc]
           info << sprintf("# %-#{max_size}.#{max_size}s<tt>%s</tt>", "*#{col_name}*::", attrs.unshift(col_type).join(", ")).rstrip + "\n"
         elsif options[:format_yard]
--- a/spec/lib/annotate/annotate_models_spec.rb
+++ b/spec/lib/annotate/annotate_models_spec.rb
@@ -1096,6 +1096,33 @@ describe AnnotateModels do
                   end
                 end
 
+                context 'when columns have multiline comments' do
+                  let :columns do
+                    [
+                      mock_column(:id,         :integer, limit: 8,  comment: 'ID'),
+                      mock_column(:notes,      :text,    limit: 55, comment: "Notes.\nMay include things like notes."),
+                      mock_column(:no_comment, :text,    limit: 20, comment: nil)
+                    ]
+                  end
+
+                  let :expected_result do
+                    <<~EOS
+                      # Schema Info
+                      #
+                      # Table name: users
+                      #
+                      #  id(ID)                                       :integer          not null, primary key
+                      #  notes(Notes.\\nMay include things like notes.):text(55)         not null
+                      #  no_comment                                   :text(20)         not null
+                      #
+                    EOS
+                  end
+
+                  it 'works with option "with_comment"' do
+                    is_expected.to eq expected_result
+                  end
+                end
+
                 context 'when geometry columns are included' do
                   let :columns do
                     [


lib/annotate/annotate_models.rb
```diff
@@ -290,11 +290,14 @@ def get_schema_info(klass, header, options = {})
             end
           end
         end
+
+        # Generate the column name optionally including column comments
         col_name = if with_comments?(klass, options) && col.comment
```
tmr08c: I considered extracting this block into a method (something like `sanitized_column_name` or `column_name_presenter`), but it didn't seem to follow current coding conventions.

lib/annotate/annotate_models.rb
```diff
@@ -290,11 +290,14 @@ def get_schema_info(klass, header, options = {})
...
-                     "#{col.name}(#{col.comment})"
+                     "#{col.name}(#{col.comment.gsub(/\n/, "\\n")})"
```
tmr08c: I wanted to find something more general and elegant than `gsub`, but didn't find a good option. I'd be interested if there are other characters that would be concerning (`\t`?) that would make sense to check for and escape as well.

drwl:
Woah neat, I didn't know [column comments](https://blog.bigbinary.com/2016/06/21/rails-5-supports-adding-comments-migrations.html) were a thing.

lib/annotate/annotate_models.rb
drwl: Maybe there's a library for sanitizing whitespace? It might be worth looking into.

lib/annotate/annotate_models.rb
drwl: Up to you how you want to handle this. Rubocop (currently) complains a lot about the project. We can always add a skip for it while it's being refactored.

drwl:
@tmr08c could you rebase and add comments to make the rubocop step part of CI pass? Happy to merge in after.

lib/annotate/annotate_models.rb
tmr08c: #791 should have resolved this

lib/annotate/annotate_models.rb
tmr08c: I looked a bit, but didn't find anything that stuck out. [`String#dump`](https://ruby-doc.org/core-2.6/String.html#method-i-dump) seemed promising, but I believe it escaped UTF characters that were expected to work.

I could make this more generic and escape any `\` with `\\` if that seems safer.

Issue #720, opened 2020-01-13
[Feature request] attribute YARD documentation for solargraph support


Samsinite:
Hi, love this gem, been using it for years.

Recently, I started using the [solargraph](https://solargraph.org/) ruby language server in my editor, which uses YARD documentation to provide intellisense like features for ruby projects. Is there any interest in adding support for this to annotate_models? If not, do you have any recommendations on the best approach to build an extension to this gem to assist with generating that documentation?

Cheers!

tvallois:
Hi,

I'm currently working on an implementation of this but i don't know if there is any interest here in adding yard support...

tvallois:
I cannot get my current implementation of this library in my gemfile using 
```ruby
gem 'annotate', '3.0.3', :git => 'https://github.com/tvallois/annotate_models'
```
The error is: 
```
Could not find gem 'annotate (= 3.0.3)' in https://github.com/tvallois/annotate_models (at master@111640d).
The source contains 'annotate' at: 2.7.5
```
don't understand why. Any idea?

Samsinite:
@tvallois would you be interested in assistance if support for documentation can be added?

Samsinite:
Ah I may have mis-understood, it looks like you meant that you are working on adding support for yard, but you weren't sure if the maintainers were interested in adding yard support.

tvallois:
hi @Samsinite,
there is already a pull request on this https://github.com/ctran/annotate_models/pull/724
Yeah we can collaborate to improve this.

tvallois:
I've tried to use the develop version of this project in our rails app using YARD but i have an error 
```
Unable to annotate app/models/preview.rb: undefined method `map_col_type_to_ruby_classes' for AnnotateModels:Module
```

Any idea @Samsinite @drwl ?

drwl:
@tvallois Can you run it with `--trace`?

tvallois:
```
tvallois@tvallois-G3-3579:~/enlaps/enlaps-web$ bundle exec bin/rake db:migrate --trace
Running via Spring preloader in process 14460
** Invoke db:migrate (first_time)
** Invoke db:load_config (first_time)
** Invoke environment (first_time)
** Execute environment
** Execute db:load_config
** Execute db:migrate
** Invoke db:_dump (first_time)
** Execute db:_dump
** Invoke db:schema:dump (first_time)
** Invoke db:load_config 
** Execute db:schema:dump
** Invoke set_annotation_options (first_time)
[...]

PR #784, opened 2020-03-24
[WIP] Fix updating annotation when formatted for Yard docs 

### LINKED ISSUE #780: Yard doc style annotations do not update after migrations
After generating annotations using the Yard doc format, running migrations does not update the annotations.

## Commands

```
bundle exec annotate --models -f yard
```

Generates:

```ruby
# == Schema Information
#
# Table name: users
#
# @!attribute id
#   @return []
# @!attribute name
#   @return [String]
# @!attribute created_at
#   @return [Time]
# @!attribute updated_at
#   @return [Time]
#
class User < ApplicationRecord
end
```

Generate a new migration and migrate:

```
> rails g migration add_email_to_users email:string
> rails db:migrate
== 20200322114415 AddEmailToUsers: migrating ==================================
-- add_column(:users, :email, :string)
   -> 0.0019s
== 20200322114415 AddEmailToUsers: migrated (0.0020s) =========================
```

Manually run annotations, but annotations are not updated:

```
› bundle exec annotate --models -f yard        
Model files unchanged.
```

## Version

- annotate version: 3.1.1
- rails version: 6.0.2.2
- ruby version: 2.6.5

## Notes

I believe the source of the problem is the regular expression used to extract columns from the annotation blob: 

https://github.com/ctran/annotate_models/blob/111640d077836ced23cfffb882c93e40968ab9d9/lib/annotate/annotate_models.rb#L510

tmr08c:
This is a WIP attempt to resolve #780  

I'd be interested in feedback on the direction of the approach. 

TODO:

- [ ] Additional tests for updating Yard docs
  - [ ]  With Yard docs, the indexes are formatted differently
  - [ ] Testing column changes
  - [ ] Trying to more robustly test the regular expression
- [ ] Add similar support for RDoc syntax

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -504,6 +504,9 @@ module AnnotateModels
     #  :force<Symbol>:: whether to update the file even if it doesn't seem to need it.
     #  :position_in_*<Symbol>:: where to place the annotated section in fixture or model file,
     #                           :before, :top, :after or :bottom. Default is :before.
+    #  :format_rdoc<Boolean>:: whether to format annotations using RDoc syntax
+    #  :format_yard<Boolean>:: whether to format annotations using Yard syntax
+    #  :format_bare<Boolean>:: whether to format annotations using default, bare syntax
     #
     def annotate_one_file(file_name, info_block, position, options = {})
       return false unless File.exist?(file_name)
@@ -515,7 +518,7 @@ module AnnotateModels
       old_header = old_content.match(header_pattern).to_s
       new_header = info_block.match(header_pattern).to_s
 
-      column_pattern = /^#[\t ]+[\w\*\.`]+[\t ]+.+$/
+      column_pattern = column_pattern_for(format_from(options))
       old_columns = old_header && old_header.scan(column_pattern).sort
       new_columns = new_header && new_header.scan(column_pattern).sort
 
@@ -996,6 +999,40 @@ module AnnotateModels
         foreign_column_name
       ]
     end
+
+    ##
+    # The format of documentation to use based on the options
+    #
+    # @param [Hash] options
+    # @option options [Boolean] :format_rdoc
+    # @option options [Boolean] :format_yard
+    # @option options [Boolean] :format_bare
+    # @return [Symbol]
+    def format_from(options)
+      if options[:format_rdoc]
+        :rdoc
+      elsif options[:format_yard]
+        :yard
+      else
+        :bare
+      end
+    end
+
+    ##
+    # Regular expression pattern used to search for a column's annotation
+    #
+    # @param [Symbol] format
+    # @return [RegularExpression]
+    def column_pattern_for(format)
+      case format
+      when :yard then /^#[\t ]+@[!]?[\w]+[\s]+.*/
+      # Default behavior matches the column pattern used by`:bare`. Instead
+      # of explicitly matching `:bare` `else` is used as a fallback to match
+      # original functionality.
+      else /^#[\t ]+[\w\*\.`]+[\t ]+.+$/
+      end
+    end
+
   end
 
   class BadModelFileError < LoadError
--- a/spec/lib/annotate/annotate_models_spec.rb
+++ b/spec/lib/annotate/annotate_models_spec.rb
@@ -2521,6 +2521,39 @@ describe AnnotateModels do
           expect(File.read(@model_file_name)).to eq("#{@schema_info}#{@file_content}")
         end
       end
+
+      context 'adding a new field' do
+        let(:class_name) { :users }
+        let(:primary_key) { :id }
+        let(:original_columns) do
+          [
+            mock_column(primary_key, :integer),
+            mock_column(:name, :string)
+          ]
+        end
+
+        before do
+          klass = mock_class(class_name, primary_key, original_columns)
+          @schema_info = AnnotateModels.get_schema_info(klass, '== Schema Info', options)
+          puts "Format yard?: #{options[:format_yard]}"
+          annotate_one_file(options)
+        end
+
+        context 'when option "format_yard" is true' do
+          let :options do
+            { format_yard: true }
+          end
+
+          it 'update the fields list to include the new column' do
+            new_column_list = original_columns + [ mock_column(:new_column, :string) ]
+            klass = mock_class(class_name, primary_key, new_column_list)
+            @schema_info = AnnotateModels.get_schema_info(klass, '== Schema Info', options)
+            annotate_one_file(options)
+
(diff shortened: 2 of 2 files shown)


lib/annotate/annotate_models.rb
```diff
@@ -362,7 +366,7 @@ def annotate_one_file(file_name, info_block, position, options = {})
       old_header = old_content.match(header_pattern).to_s
       new_header = info_block.match(header_pattern).to_s
 
-      column_pattern = /^#[\t ]+[\w\*\.`]+[\t ]+.+$/
+      column_pattern = /^#[\t ]+[\w\*\.\`\@\!\:]+[\t ]+.+$/
```
tmr08c, line 369: I was originally going to break this regular expression up into something like:

```ruby
def column_pattern_for(format_type)
  case format_type
   when :yard then //
   when :rdoc then //
    # etc...
    end
end
```

By the time I go to the end, I saw that the original regular expression could be updated and cover all cases. 

I could see an advantage to breaking this up, but didn't want to go too far with refactoring and delay sharing this fix.

tmr08c:
@tvallois since you added Yard support (#724) and @Samsinite since you were interested in YARD support (#720), I figured I'd ping you both to see if you had any thoughts around this change or experienced this issue.

lib/annotate/annotate_models.rb
drwl: @tmr08c if you're still interested in getting this patch in, I think there's value into breaking it up. The column pattern regex is hard to understand already.

lib/annotate/annotate_models.rb
```diff
@@ -908,6 +913,58 @@ def get_attributes(column, column_type, klass, options)
...
+    # the format of the annotations being made
+    #
+    # == Returns:
+    # Regular expression
+    #
+    # @param [Symbol] symbol representation of the format being used for
+    #   annotations
+    def column_pattern_for(format_type)
+      case format_type
+      when :markdown then /^#\s+\*{2}`\w+`\*{2}(?:\s+\|\s`.*`)*$/
```
tmr08c, line 927: I went for stricter matches. I'm not sure if it would make sense to be a bit looser and do something like:

```
/^#\s+\*{2}`\w+.*$/
```

that would catch 

```
# **column_name** ...
```

Currently, I'm checking for the fact there _may_ be additional columns in the markdown table

lib/annotate/annotate_models.rb
```diff
@@ -908,6 +913,58 @@ def get_attributes(column, column_type, klass, options)
...
+      when :rdoc then
+        field_name_regex = /^#\s+\*\w+\*\:{2}/
+        field_info_regex = /\s+\<tt\>.*\<\/tt\>$/
```
tmr08c, line 930: Similar to above, rather than checking the `<tt>...</tt>` this could all probably be `.*$`

lib/annotate/annotate_models.rb
```diff
@@ -908,6 +913,58 @@ def get_attributes(column, column_type, klass, options)
...
+
+        /#{field_name_regex}#{field_info_regex}/
+      when :yard
+        attribute_regex = /^#\s@!attribute\s\w+$/
+        return_regex = /^#\s+@return\s+\[\w+\]$/
+
+        /#{attribute_regex}|#{return_regex}/
+      else # :bare/default
+        /^#[\t ]+[\w\*\.\`\@\!\:]+[\t ]+.+$/
```
tmr08c, line 939: This is the catch-all regular expression. I would like to move it to something more specific to the `bare` format.

My initial implementation broke something for foreign keys, so I will need to see how that is handled for the various formats.

lib/annotate/annotate_models.rb
tmr08c: Thanks for the review and feedback @drwl! Working on updating this PR and adding some additional tests.

lib/annotate/annotate_models.rb
tmr08c: After revisiting this PR, I'm not sure how I feel about this. I'm sure if we just grabbing all lines in the annotation section would end up being a problem. 

The previous implementation seemed to sort of be checking that the comment had an indication that it was the "column section" (e.g., `# col_name` or `# * col_name`).  This PR goes a bit further and does some stricter checking based on the format.

If we pulled in any non-empty comment line, it seems it _shouldn't_ cause a problem since we are essentially [checking for _any_ difference](https://github.com/ctran/annotate_models/pull/784/files#diff-2741ac3b166ed72003c60347ac3e3586R388). If we pulled in non-column comments, we'd get them in both `old_columns` and `new_columns`, so it shouldn't be flagged as a change. 

I _think_ the situation that we are avoiding by doing it this way is if there are non-annotation comments in the annotation block since it looks like the `header_pattern` grabs all groups comments after the `Table name:` comment.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -2526,44 +2526,256 @@ def annotate_one_file(options = {})
...
+          klass = mock_class(class_name, primary_key, original_columns)
+          @schema_info = AnnotateModels.get_schema_info(klass, '== Schema Info', options)
+          annotate_one_file(options)
+
+          # confirm we initialized annotaions in file before checking for changes
+          expect(@schema_info).not_to be_empty
+          expect(File.read(@model_file_name)).to eq("#{@schema_info}#{@file_content}")
+        end
+
+        context 'when option "format_bare" is true' do
```
tmr08c, line 2629: I started going down the [shared examples](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) route to reduce the noise here. This snowballed a bit since there are additional places where checking for all formats would be useful. I have [a WIP branch](https://github.com/ctran/annotate_models/compare/develop...tmr08c:tmr08c-shared-examples-model-annotation#diff-d35383c256402bb04c05c0f176b1d182) that I'd like to pick up if it looks like the right direction. 

This PR ended up being bigger than I expected, so I thought it would be best to separate additional changes and conversations.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -2526,44 +2526,256 @@ def annotate_one_file(options = {})
...
+                                 updated_foreign_keys)
+
+              @schema_info = AnnotateModels.get_schema_info(klass, '== Schema Info', options)
+              annotate_one_file(options)
+
+              expect(File.read(@model_file_name)).to eq("#{@schema_info}#{@file_content}")
+            end
+          end
+
+          context 'when option "format_markdown" is true' do
```
tmr08c, line 2590: There didn't appear to be a test for the markdown format, so this adds a test for `format_markdown` and extracts some of the shared set up into `before`s and `let`s.

PR #790, opened 2020-04-02
Refactor AnnotateRoutes by adding AnnotateRoutes::HeaderGenerator

nard-tech:
I noticed that `AnnotateRoutes` can be more maintainable by refactoring.

I am planning to refactor `AnnotateRoutes` in this order.
*   separate logic of `AnnotateRoutes` into `AnnotateRoutes::HeaderGenerator`.
*   add methods to `AnnotateRoutes::HeaderGenerator` and refactor methods.
*   add `AnnotateRoutes::AnnotationProcessor` and `AnnotateRoutes::RemovalProcessor`

The final goal of this refactoring is as follows.
* https://github.com/nard-tech/annotate_models/blob/feature/refactor_annotate_routes/processors/lib/annotate/annotate_routes.rb
* https://github.com/nard-tech/annotate_models/tree/feature/refactor_annotate_routes/processors/lib/annotate/annotate_routes

So in the first I added `AnnotateRoutes::HeaderGenerator` in order to separate logic of `AnnotateRoutes` in this PR.

When refactor of `AnnotateRoutes` is finished, I would like to refactor `AnnotateModels` in a like way.


tmr08c:
I'm not a maintainer on the project, but if it helps make reviewing easier for others, it looks like this does what you described in your first point - separate logic of `AnnotateRoutes` into `AnnotateRoutes::HeaderGenerator` 👍

PR #803, opened 2020-05-09
load :set_annotation_options before running :annotate_routes

### LINKED ISSUE #761: Unable to annotate routes at the bottom while it works correctly for Models
Unable to `annotate_routes` at the bottom of the file. This is working for `models` as expected. 

## Commands

```
$ rails annotate_routes
```

## Configuration
```
if Rails.env.development?
  require 'annotate'
  task :set_annotation_options do
    # same name.
    Annotate.set_defaults(
      'active_admin'                => 'false',
      'additional_file_patterns'    => [],
      'routes'                      => 'true',
      'models'                      => 'true',
      'position_in_routes'          => 'after',
      'position_in_class'           => 'after',
      'position_in_test'            => 'after',
      'position_in_fixture'         => 'after',
      'position_in_factory'         => 'after',
```

## Versions

- 3.1.0
- 6.0.2.1
- 2.6.5

javawizard:
Make `annotate_routes` depend on `set_annotation_options` just as `annotate_models` does.

Fixes #761.

I copied the `ENV['is_cli']` check from `annotate_models.rake`, but I don't know what its purpose is. Feel free to tell me to get rid of it. I also moved the `annotate_lib = ...` bit to the top of the file to keep `annotate_routes.rake` consistent with `annotate_models.rake`.

### DIFF
--- a/lib/tasks/annotate_routes.rake
+++ b/lib/tasks/annotate_routes.rake
@@ -1,6 +1,12 @@
+annotate_lib = File.expand_path(File.dirname(File.dirname(__FILE__)))
+
+unless ENV['is_cli']
+  task :set_annotation_options
+  task annotate_routes: :set_annotation_options
+end
+
 desc "Adds the route map to routes.rb"
 task :annotate_routes => :environment do
-  annotate_lib = File.expand_path(File.dirname(File.dirname(__FILE__)))
   require "#{annotate_lib}/annotate/annotate_routes"
 
   options={}


drwl:
Hi @javawizard, I'm not sure what `ENV['is_cli']` does either. Currently it looks like dead code as we're not using it in anywhere in the project.

drwl:
For the time being, let's keep it as is, and we can remove it in a future release.

Issue #816, opened 2020-06-17
annotate is not updated when adding a column(with comment)


umanoda:
Annotate is not updated when adding a new commented column to a table that has commented columns.

## Commands

```rb
$ cat > app/models/user.rb
class User <  ApplicationRecord
end

$ cat > db/migrate/20200617075230_create_user.rb   
class CreateUser < ActiveRecord::Migration[5.2]
  def change
    create_table :users do |t|
      t.string :name, comment: 'Name'
      t.timestamps
    end
  end
end

$ bin/rails db:migrate 

$ cat >  db/migrate/20200617075237_create_user.rb 
class AddAddressToUser < ActiveRecord::Migration[5.2]
  def change
    add_column :hoges, :age, :integer, after: :name, comment: 'Age'
  end
end

$ bin/rails db:migrate 
```

got / expected(- line)

```diff
  # == Schema Information
  #
  # Table name: hoges
  #
  #  id           :bigint(8)        not null, primary key
  #  name(Name)   :string(255)
- #  age(Age)     :integer
  #  created_at   :datetime         not null
  #  updated_at   :datetime         not null
  #
  class User <  ApplicationRecord 
  end
```

## Version

- annotate version: 3.1.1
- rails version: rails-5.2.3
- ruby version: 2.6.5p114

PR #817, opened 2020-06-18
Fixed columns difference detection

umanoda:
See also https://github.com/ctran/annotate_models/issues/816

When commented columns was include on annotate, a new commented column was not added.The difference could not be detected because `(comment)` is included after the column name.

Fixed pattern match to detect differences even when `with_comment` option is enabled.


umanoda:
It's been a long time, so I'll close it.

PR #838, opened 2020-09-24
Put Comments on New Line for Bare Formatting

deiden26:
Prevent unusably spaced formatting that occurs when comments are the
length of normal sentences or even full paragraphs.


ctran:
Hi, I'm not sure I understand what's being done here.

apotema:
I think that he wants to change from this

\#  tide_member(True if the user is a permanent tide member.) :boolean          default(FALSE), not null

to this 

\#  True if the user is a permanent tide member.
\#  tide_member :boolean          default(FALSE), not null

That would be very helpful in reducing the length of the annotations.

Issue #758, opened 2020-02-19
no implicit conversion of Pathname into String


eni9889:
Describe your problem here.

## Commands

```
bundle exec rake annotate_models
```

example model below. Also, primary keys are uuid in case that matters
```
class Site < ApplicationRecord
  belongs_to :user
  validates :anchor_link, url: { allow_blank: true, allow_nil: true }

  after_create_commit :post_creation_callback

  def post_creation_callback
    Seo::ChooseAnchorTextJob.perform_later(self.class.to_s, id)
  end

  after_save do
    if saved_change_to_anchor_text? || saved_change_to_anchor_link?
      Seo::SyncAnchorTextJob.set(wait: 1.minutes).perform_later(self.class.to_s, self.id) rescue nil
    end
  end
end

```

## Version

- annotate version: 3.0.3
- rails version: 5.2.3
- ruby version: 2.5.1p57

a0s:
I've got same problem with namespaced models. Structure of files are:

```
/app/models/rosstat.rb                     # empty module Rosstat
/app/models/rosstat/people_income.rb       # empty module PeopleIncome
/app/models/rosstat/people_income/file.rb  # class Rosstat::PeopleIncome::File < ApplicationRecord
/app/models/rosstat/people_income/value.rb # class Rosstat::PeopleIncome::Value < ApplicationRecord
```

## Error

```
Unable to annotate app/models/rosstat/people_income/value.rb: no implicit conversion of Pathname into String
Unable to annotate app/models/rosstat/people_income/file.rb: no implicit conversion of Pathname into String
Unable to annotate app/models/rosstat/people_income.rb: no implicit conversion of Pathname into String
```

## Settings

```ruby
Annotate.set_defaults(
      'active_admin'                => 'false',
      'additional_file_patterns'    =>  [],
      'routes'                      => 'false',
      'models'                      => 'true',
      'position_in_routes'          => 'before',
      'position_in_class'           => 'before',
      'position_in_test'            => 'before',
      'position_in_fixture'         => 'before',
      'position_in_factory'         => 'before',
      'position_in_serializer'      => 'before',
      'show_foreign_keys'           => 'true',
      'show_complete_foreign_keys'  => 'false',
      'show_indexes'                => 'true',
      'simple_indexes'              => 'false',
      'model_dir'                   => 'app/models',
      'root_dir'                    => '',
      'include_version'             => 'false',
      'require'                     => '',
      'exclude_tests'               => 'true',
[...]

PR #848, opened 2021-01-26
Allow `get_loaded_model` to succeed when `$LOAD_PATH` contains non-string values

Hamms:
As currently implemented, `get_loaded_model` inspects the `$LOAD_PATH` global for path values when trying to find the path for a model file.  This would be fine, except that variable is affected by userspace, which means that it will sometimes contain non-string values, often Pathnames.  To avoid responding with the error `Unable to annotate #{model_path}: no implicit conversion of Pathname into String` in this situation, we simply add an explicit `to_s` call before performing string-specific operations.

Fixes https://github.com/ctran/annotate_models/issues/758


bopm:
@ctran any news on this one?

ctran:
Thanks!

PR #849, opened 2021-01-27
Fix retrieve_indexes_from_table when indexes is empty and base table does not exist

jjowdy:
Some tables may have a table_name_prefix but no indexes. Previous versions of
the code would strip the prefix and look for indexes on the resulting table name,
which likely would not exist. This causes DB errors, at least in MySQL. So now
check if the table with the derived name exists first before trying to show its indexes.

Example error otherwise:
```
myservice:development [10] pry(main)> ActiveRecord::Base.connection.indexes('garbage')
Mysql2::Error: Table 'myservice_development.garbage' doesn't exist: SHOW KEYS FROM `garbage`
ActiveRecord::StatementInvalid: Mysql2::Error: Table 'myservice_development.garbage' doesn't exist: SHOW KEYS FROM `garbage`
```

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -124,7 +124,11 @@ module AnnotateModels
 
       # Try to search the table without prefix
       table_name_without_prefix = table_name.to_s.sub(klass.table_name_prefix, '')
-      klass.connection.indexes(table_name_without_prefix)
+      if klass.connection.table_exists?(table_name_without_prefix)
+        klass.connection.indexes(table_name_without_prefix)
+      else
+        []
+      end
     end
 
     # Use the column information in an ActiveRecord class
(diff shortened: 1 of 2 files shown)


ctran (approved):
Thanks!

rajyan:
Is there any chance that this change being merged?
I've encountered this bug recently, and it's very hard to notice the cause from the error message...

jjowdy:
> Is there any chance that this change being merged?
> I've encountered this bug recently, and it's very hard to notice the cause from the error message...

I don't think I have write permissions so I can't seem do that myself. @ctran would you have a chance to do this?

ctran:
Something is strange about this PR.  I don't see the usual checks being triggered on this one.

ctran:
Can you rebate and resolve the conflict?

jjowdy:
> Can you rebate and resolve the conflict?

Sure. (That was a bit of a context switch: I no longer have access to the system I made the fix on and hadn't pushed to github in a while.) I wrapped the new spec context with `rubocop:disable RSpec/NestedGroups` though because I won't have time to address that.

ctran:
Thanks!

rajyan:
Thank you both so much for taking care of this issue!

PR #793, opened 2020-04-06
Add AnnotateRoutes::AnnotationProcessor and AnnotateRoutes::RemovalProcessor

nard-tech:
cf. #790, #792

In order to refactor AnnoateRoutes, I added `AnnotateRoutes::AnnotationProcessor` and `AnnotateRoutes::RemovalProcessor`.

Refactoring AnnotateRoutes is finished in this PR and I would like to refactor AnnotateModels in a like way.

PR #851, opened 2021-02-06
feat: Annotation support for table comment

TomohikoSato:
Add annotation support for table comment.

It would be useful for many people because table comment is recommended as described in the rails guide.

https://guides.rubyonrails.org/active_record_migrations.html
>  It's highly recommended to specify comments in migrations for applications with large databases as it helps people to understand data model and generate documentation.


ctran:
Thanks.  Could you post an example with both table and column comments?

TomohikoSato:
resolve CI error by cherry-pick 5cc83c6 , same as https://github.com/ctran/annotate_models/pull/793#issuecomment-814690403

TomohikoSato:
@ctran 

>  Could you post an example with both table and column comments?

OK. added 8195810

I resolve CI error by cherry-picking 5cc83c6 , same as https://github.com/ctran/annotate_models/pull/793#issuecomment-814690403

PR #868, opened 2021-04-23
Add support for annotating check constraints

lovro-bikic:
Rails added support for check constraints (https://github.com/rails/rails/pull/31323), so I thought it would be nice if we could annotate them in models.

This PR adds annotation of check constraints with an option to disable/enable annotation. Most of the work done in this PR is based off of existing implementation for annotating indexes and foreign keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ you can do so with a simple environment variable, instead of editing the
         -a, --active-admin               Annotate active_admin models
         -v, --version                    Show the current version of this gem
         -m, --show-migration             Include the migration version number in the annotation
+        -c, --show-check-constraints     List the table's check constraints in the annotation
         -k, --show-foreign-keys          List the table's foreign key constraints in the annotation
             --ck, --complete-foreign-keys
                                          Complete foreign key names in the annotation
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -178,6 +178,10 @@ module AnnotateModels
         info << get_foreign_key_info(klass, options)
       end
 
+      if options[:show_check_constraints] && klass.table_exists?
+        info << get_check_constraint_info(klass, options)
+      end
+
       info << get_schema_footer_text(klass, options)
     end
 
@@ -350,6 +354,31 @@ module AnnotateModels
       end
 
       fk_info
+    end
+
+    def get_check_constraint_info(klass, options = {})
+      cc_info = if options[:format_markdown]
+                  "#\n# ### Check Constraints\n#\n"
+                else
+                  "#\n# Check Constraints\n#\n"
+                end
+
+      return '' unless klass.connection.respond_to?(:supports_check_constraints?) &&
+        klass.connection.supports_check_constraints? && klass.connection.respond_to?(:check_constraints)
+
+      check_constraints = klass.connection.check_constraints(klass.table_name)
+      return '' if check_constraints.empty?
+
+      max_size = check_constraints.map { |check_constraint| check_constraint.name.size }.max + 1
+      check_constraints.sort_by(&:name).each do |check_constraint|
+        cc_info << if options[:format_markdown]
+                     sprintf("# * `%s`: `(%s)`\n", check_constraint.name, check_constraint.expression)
+                   else
+                     sprintf("#  %-#{max_size}.#{max_size}s (%s)\n", check_constraint.name, check_constraint.expression)
+                   end
+      end
+
+      cc_info
     end
 
     # Add a schema block to a file. If the file already contains
--- a/lib/annotate/constants.rb
+++ b/lib/annotate/constants.rb
@@ -18,7 +18,8 @@ module Annotate
       :trace, :timestamp, :exclude_serializers, :classified_sort,
       :show_foreign_keys, :show_complete_foreign_keys,
       :exclude_scaffolds, :exclude_controllers, :exclude_helpers,
-      :exclude_sti_subclasses, :ignore_unknown_models, :with_comment
+      :exclude_sti_subclasses, :ignore_unknown_models, :with_comment,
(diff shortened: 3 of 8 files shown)


Eusebius1920:
Thanks for sharing. This should get merged!

deecewan:
@ctran any change we can get this merged?

bruno-:
Here's a 2022 bump to this issue.

`check_constraint`s are great and it would be great having them in the annotations.

dtcristo:
Hi @lovro-bikic,

First of all, I love the new feature ❤️.

We've started using your implementation and noticed it has issues with check constraints that go across multiple lines.

For example, say I have the following constraint:
```sql
CONSTRAINT foo_not_null_when_active CHECK (
CASE
    WHEN ((status)::text = 'active'::text) THEN (foo IS NOT NULL)
    ELSE true
END)
```

Currently this results in the following annotation:
```
# Check Constraints
#
#  foo_not_null_when_active  ()
#
```

Not sure how we'd want this to be formatted in the annotation? Perhaps we can keep the wrapping. Maybe something like this?

```
# Check Constraints
#
#  foo_not_null_when_active  (CASE
#                                WHEN ((status)::text = 'active'::text) THEN (foo IS NOT NULL)
#                                ELSE true
#                            END)
#
```

Alternatively could `.squish` it and print:
```
# Check Constraints
#
#  foo_not_null_when_active  (CASE WHEN ((status)::text = 'active'::text) THEN (foo IS NOT NULL) ELSE true END)
#
```

lovro-bikic:
@dtcristo, thanks for the feedback!

I would go with multiline expressions squished into a single line, to be consistent with the rest of the annotations. I pushed a commit for it: https://github.com/ctran/annotate_models/pull/868/commits/b850e9b52d17d9b99de09f53bfb0d26e173f4be3.

Side note: CI is failing, though the failures are unrelated to this branch. `develop` branch currently has the same [failures](https://github.com/ctran/annotate_models/runs/5292042822?check_suite_focus=true).

dtcristo:
Hi @lovro-bikic 👋, I've had a play with your `.squish` changes in my Rails 6.1 application and running in to some issues.

So it turns out there is a bug in Rails 6.1 (since [fixed here](https://github.com/rails/rails/pull/43963) for Rails 7 only). The bug means that multi-line constraints are coming through as `nil` from the connection for PostgreSQL.

So your line:
```
check_constraint.expression.squish
```
Will error because `check_constraint.expression` is `nil` for these multi-line constraints. **So at the very least we might want `&.squish`** so we don't raise an exception in this case and keep the behaviour of printing nothing for users with older versions of Rails (like me).

However, something else I noticed with the change [here](https://github.com/rails/rails/pull/43963), is that it will introduce an extra later of parenthesis around the check constraint for normal (one line) constraints.

So a constraint that was previously printing as:
```
alive    (age < 150)
```

Will now print
```
alive    ((age < 150))
```

However with this change a multi-line constraint will correctly print with only one layer of parenthesis.

Not sure if this is an issue, but I just wanted to point it out? Potentially if you see wrapping parenthesis you could strip them?

lovro-bikic:
Thanks for pointing out the issue regarding missing expressions, it should be fixed with this commit: https://github.com/ctran/annotate_models/pull/868/commits/6449e7d701f3fb72b417ba6e799b57d9dc7d5b4e.

> However, something else I noticed with the change https://github.com/rails/rails/pull/43963, is that it will introduce an extra later of parenthesis around the check constraint for normal (one line) constraints.

I couldn't reproduce this. I tried it out on Rails versions 6.1.5.1 and 7.0.2.4, and Rails in both cases output the constraint expression without the wrapping parentheses, so the annotation would always correctly have a single pair of them.

lparry:
It would be really helpful to get this merged. Is there anything we might to to help this along @ctran ?

ctran:
I need to take a look to see why the CI is failing.

rjherrera:
@ctran hi there! have you been able to look at this? We found ourselves needing to annotate constraints, and found out about this pull request, it looks like what we need 😄, anything we can do to help get this merged?

ctran:
> @ctran hi there! have you been able to look at this? We found ourselves needing to annotate constraints, and found out about this pull request, it looks like what we need 😄, anything we can do to help get this merged?

I just need help getting the CI build to pass :-)

lovro-bikic:
@ctran I rebased the branch but the CI is still failing due to this error:
```
  1) AnnotateModels annotating a file frozen option should abort without existing annotation when frozen: true 
     Failure/Error: expect { annotate_one_file frozen: true }.to raise_error SystemExit, /user.rb needs to be updated, but annotate was run with `--frozen`./
       expected SystemExit with message matching /user.rb needs to be updated, but annotate was run with `--frozen`./ but nothing was raised
     # ./spec/lib/annotate/annotate_models_spec.rb:2914:in `block (4 levels) in <top (required)>'
```

It seems that the test suite is non-deterministic because when I run `bundle exec rspec` locally sometimes it passes and sometimes it fails (either with the above error or some other errors), depending on the RSpec seed. I found that running with the seed `43289` (`bundle exec rspec --seed 43289`) makes the suite pass.

I'm not sure how to proceed here because these errors fall outside the scope of this PR, please advise.

lovro-bikic:
@ctran I invested a bit of time in researching the flaky test suite and found what causes these issues. Here is the PR which should fix all of them: https://github.com/ctran/annotate_models/pull/980. After we merge that, I will rebase this branch.

ctran:
@lovro-bikic I merged #980

lovro-bikic:
Thanks! I rebased the branch, but the CI failed because of Rubocop. I disabled these offenses but let me know if you want me to fix them in a different way.

ctran (approved):
LGTM (we'll deal with the rubocop issues later)

PR #871, opened 2021-04-26
Fix CI

drwl:


olleolleolle:
@drwl Seems like the CI is nowadays green, and perhaps this can be closed? Thanks for working to fix annotate!

Issue #881, opened 2021-06-01
When is the next release scheduled?


sato11:
So #866 is actually already fixed in #779, right?
Also I've seen a handful of PRs merged since the last release which is a year ago. 

Now I thought it would be great if we could know if there is a plan for releasing the next version.
Could you make that happen? @ctran 

By the way thanks for the neat product 🙂 
I've been using it for a couple of years and your job has helped me a lot already. Cheers.

ctran:
I'll do a release either this weekend or next week, depending on my schedule.

sato11:
Thanks for bringing a great news 😊 I'll look forward to it!

sato11:
@ctran 
Hi, how is it going? I'm looking forward for the release and I'd like to hear the updated information about it. Is everything alright? Let me know if there is something that I can help you with. Thanks!

ctran:
This is released - https://github.com/ctran/annotate_models/releases/tag/v3.2.0

PR #883, opened 2021-06-10
`get_loaded_model_by_path` is supposed to be nil-safe

sato11:
This commit partially reverts #801, which by declaring conditional return
has turned `get_loaded_model_by_path` to a less safe method that can
return nil when its condition is not met.

Apparently the very same condition has been brought to `annotate_model_file`
by #774, which seems to cover the "bug" insisted in #801 as well.

On the other hand #801 has brought an inconvenient behaviour as well:
whenever a non-activerecord model file is found, `get_loaded_model_by_path`
returns nil, which leads to raising `BadModelFileError` and ends up
printing a bunch of "Unable to annotate ..." messages.

Now it seems tests added by #801 are running right and I do not find
a problem restoring the previous behaviour and turn it nil-safe again.

---

Say we have a file like `app/models/foo/bar.rb` which is NOT a subclass of ActiveRecord::Base:

```ruby
class Foo::Bar
end
```

We do not want this to get annotated but the current code does give it a try, ending up showing a bunch of error messages like this:

```
Unable to annotate app/models/foo/bar.rb: file doesn't contain a valid model class
Unable to annotate app/models/foo/baz.rb: file doesn't contain a valid model class
...
```

With this change the behaviour in v3.1.1 is restored and these un-informational errors are gone.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -618,9 +618,7 @@ module AnnotateModels
 
     # Retrieve loaded model class by path to the file where it's supposed to be defined.
     def get_loaded_model_by_path(model_path)
-      klass = ActiveSupport::Inflector.constantize(ActiveSupport::Inflector.camelize(model_path))
-
-      klass if klass.is_a?(Class) && klass < ActiveRecord::Base
+      ActiveSupport::Inflector.constantize(ActiveSupport::Inflector.camelize(model_path))
     rescue StandardError, LoadError
       # Revert to the old way but it is not really robust
       ObjectSpace.each_object(::Class)


sato11:
@ctran 
Will you take a look at this before proceeding with v3.1.2? (#881)
The CI's failed but I guess it's a config issue since all the PRs have failed too since mid-April?

ctran:
I'll take a look at the CI configuration.  You want this to be merged for 3.1.2 right?

sato11:
Yes I do since v3.1.2 would introduce breaking change if shipped without this. Thanks for the quick action! 🙌

PR #892, opened 2021-07-29
Fix conflicting annotations for multiple-database scenario

### LINKED ISSUE #891: Incorrect annotation of factory files when different models share the same table name in different databases
Given a Rails application with multiple databases, tables in two databases with the same table name, and corresponding models with different names, some files get annotated with the wrong table's schema.

The issue is that some of the patterns include `%TABLE_NAME%`, which will be identical between the two tables, even though they have different model names.

I'll be following up with a PR shortly to address the issue.

afn:
Given a Rails application with multiple databases, tables in two databases with the same table name, and corresponding models with different names, some files get annotated with the wrong table's schema.

The issue is that some of the patterns include `%TABLE_NAME%`, which will be identical between the two tables, even though they have different model names.

This fixes the issue by eliminating the use of the table name when the model is not connected to the primary database.

Fixes #891.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -480,7 +480,7 @@ module AnnotateModels
         klass.reset_column_information
         info = get_schema_info(klass, header, options)
         model_name = klass.name.underscore
-        table_name = klass.table_name
+        table_name = klass.table_name if klass.connection_specification_name == ActiveRecord::Base.name
         model_file_name = File.join(file)
         annotated = []
 
@@ -702,7 +702,7 @@ module AnnotateModels
           klass = get_model_class(file)
           if klass < ActiveRecord::Base && !klass.abstract_class?
             model_name = klass.name.underscore
-            table_name = klass.table_name
+            table_name = klass.table_name if klass.connection_specification_name == ActiveRecord::Base.name
             model_file_name = file
             deannotated_klass = true if remove_annotation_of_file(model_file_name, options)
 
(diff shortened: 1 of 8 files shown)


pkopac (approved):
Our app also got affected, I had to temporarily disable other databases to be able to annotate correctly.

afn:
I've just rebased this branch to `develop`. Note that this branch previously had an integration test for the new behavior, but since the integration tests were removed in #959, these tests are no longer included.

If integration tests are reintroduced upstream, they're in this commit: 385c293a0dbc2f822b77817de79c527c68b30d19.

drwl:
@afn sorry to hijack the thread a bit, but if you're interested I could get this added to my fork https://github.com/drwl/annotaterb. Since I'm still on the list of maintainers for this gem I get pinged about activity by Github every now and then.

I would just need instructions on setting up a proof of concept Rails app with the case you're talking about (multiple databases + models with same table name).

afn:
Thanks @drwl!

I've pushed an example app to https://github.com/cygnuseducation/annotate-test-app. This test case doesn't actually end up incorrectly annotating the files, but it does incorrectly detect that there are relevant changes to the annotations:

```
➜  test-app git:(main) annotaterb models
Annotating models
Annotated (2): test/factories/users.rb, test/factories/users.rb
```

See also the integration test in 385c293a0dbc2f822b77817de79c527c68b30d19.

Thanks!

drwl:
@afn would you mind if we continued discussions in https://github.com/drwl/annotaterb/pull/135?

PR #906, opened 2021-10-16
Don't write if there are no changes.

mcordingley:
This saves on disk I/O, which adds up for large projects. This should speed up model annotation, since lots of file writes can be skipped.

If this PR is acceptable, I'd much appreciate it if you could apply the `hacktoberfest-accepted` label to it.


ctran:
Sorry for the late response, please add a test.

PR #908, opened 2021-10-19
Adds typed and warn_indent as magic comments

JohnnysRibeiro:
As of right now, annotate models only considers `frozen_string_literal` and `encoding` as magic comments. This pr adds `typed` and `warn_indent` to the list.


ctran:
Could you add some tests?  Thanks.

PR #911, opened 2021-12-15
fix: support column with Unicode comment

ota42y:
When we crete column with Unicode comment, annotate gem does not update that annotation.

`\w` means a word character([a-zA-Z0-9_]) and it does not match Unicode leteers.
https://docs.ruby-lang.org/en/3.0.0/Regexp.html

Now `column_pattern` in `annotante_one_file` support
ascii characters only and doesn't match column with unicode comment.
So the columns will disappear from `old_columns` and `new_columns`.

https://github.com/ctran/annotate_models/blob/786394947c041f781df2ee0ea003e09452fa9dba/lib/annotate/annotate_models.rb#L378-L380

Even if we change that column (i.e. nullable to not null),
annotate_one_file's check ignore that column so annotation doesn't update.
(The first time create it, it works, and the column annotation is left out of date)

So I fix `column_pattern` support Unicode letter.


ctran (approved):
LGTM.  I'm not sure why the "ci" build didn't get triggered?

ota42y:
I fixed rubocop warning and these test failed occurred in develop branch so I think this change is no problem.

PR #912, opened 2021-12-16
Loosen activerecord restriction to work with rails 7

dabit:
This change allows the gem to work with Rails 7 now that it's out.

I've been using it for a while with alpha and RC versions and at least models and routes work fine.

EDIT:

~~Just included as a bonus a `rake routes` task that fixes the problems that showed up when rails renamed the task to `rails routes`~~

### DIFF
--- a/annotate.gemspec
+++ b/annotate.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
 
   s.specification_version = 4 if s.respond_to? :specification_version
   s.add_runtime_dependency(%q<rake>, '>= 10.4', '< 14.0')
-  s.add_runtime_dependency(%q<activerecord>, ['>= 3.2', '< 7.0'])
+  s.add_runtime_dependency(%q<activerecord>, ['>= 3.2', '< 7.1'])
 
   s.metadata = {
     "bug_tracker_uri" => "https://github.com/ctran/annotate_models/issues/",
--- a/annotate.gemspec
+++ b/annotate.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
 
   s.specification_version = 4 if s.respond_to? :specification_version
   s.add_runtime_dependency(%q<rake>, '>= 10.4', '< 14.0')
-  s.add_runtime_dependency(%q<activerecord>, ['>= 3.2', '< 7.1'])
+  s.add_runtime_dependency(%q<activerecord>, ['>= 3.2', '< 8.0'])
 
   s.metadata = {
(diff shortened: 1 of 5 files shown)


davidteren (approved):
LGTM

alfie-max:
@ctran can we have this merged in? I've been using this with rc versions till yesterday and everything looks good

artplan1 (approved):

thenamankumar:
@ctran requesting a quick merge for this.

Flink:
When do you think this could be merged? :)

StoneGod:
@ctran

ayoa77:
@ctran

annotate.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
 
   s.specification_version = 4 if s.respond_to? :specification_version
   s.add_runtime_dependency(%q<rake>, '>= 10.4', '< 14.0')
-  s.add_runtime_dependency(%q<activerecord>, ['>= 3.2', '< 7.0'])
+  s.add_runtime_dependency(%q<activerecord>, ['>= 3.2', '< 7.1'])
```
kriansa, line 26: Wouldn't it be better to use `8.0` instead? I think it will be using the same rationale for the existing file: by relying on the fact that a major Rails version will not be significantly incompatible with this gem.

mileszim:
@ctran any chance we can merge this in?

slayer:
Looking forward for merge )

annotate.gemspec
ctran: See https://github.com/ctran/annotate_models/pull/915

dabit: Done

ctran:
Hm, any reason the "ci" build is failing?

dabit:
I thought it was the extra rake task, but now that I've reverted we get the same result. Not happening on my computer with different versions either. Still trying to replicate.

ctran:
Appreciate the help.  I'm trying to do a release with this change included.

dabit:
No prob, whatever works!

ctran:
I confirmed the dependency change caused the ci failure.  I'm not sure why they failed though.

yatryan:
The failures could be related to Rails 7 supporting only ruby 2.7 and up? Source: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html

ctran:
This requires updates to the Gemfile and also causes the integration tests to fail.

dabit:
@ctran I think I figured it out

ctran (approved):
Thanks!

PR #920, opened 2022-01-14
Fix undefined method error when spatial column does not have `srid`

oieioi:
Fix undefined method error when spatial column does not have `srid`. This error occurs when the following gem is used.

https://github.com/sebastianiorga/activerecord-mysql2spatial-adapter
(forked from https://github.com/rgeo/activerecord-mysql2spatial-adapter)

```
Unable to annotate app/models/location.rb: undefined method `srid' for #<ActiveRecord::ConnectionAdapters::Mysql2SpatialAdapter::SpatialColumn:...
```


ctran:
Thanks!

PR #931, opened 2022-02-09
Update to rubocop 1.12.x for newer ruby support

gravitystorm:
In #927 the CI builds for 2.7 and 3.0 failed, because rubocop 0.68.x is too old to recognise them.

This PR updates rubocop to 1.12.x, which is the last version that supports ruby 2.4. That version of rubocop also supports 2.7 and 3.0, so this PR will enable #927 to work (after rebasing).

The bulk of the diff in this PR is just running `rubocop --auto-gen-config` after the rubocop update.


ctran:
Thanks!  This was updated in HEAD.

Issue #951, opened 2022-04-29
Comments get lost when generating annotations


cmer:
When there are already comments above a class definition, AnnotateModels deletes them when regenerating annotations.

Here's a short video showing the issue: https://cln.sh/Iw7NsPk7yEfmkyPWtIim

## Version

- annotate version: 3.2
- rails version: 5.2.6.2
- ruby version: 2.7.5

joevandyk:
running into this as well

alexeyr-ci1:
It works to add a blank line above the comment before annotating, but you need to find them first (I spotted it accidentally when reviewing a PR).

cmer:
Yes but obviously nobody remembers to do that on a large team :)

alexeyr-ci1:
Sorry, it's actually worse than I thought. It looked like it worked only because the comment wasn't getting regenerated. If I use `--force` (or do a schema change which forces a regeneration), then the first time eats the empty line and leaves the comment intact (so really hard to note in code review), and the next time eats the comment 🤣.

filipesperandio:
Is there a workaround for this?

drwl:
Hey folks, I believe I was able to solve this bug in my gem https://github.com/drwl/annotaterb/pull/72

ghost:
I wasn't aware of the fork. I'll definitely check it out. Thanks!

fiedl:
> It works to add a blank line above the comment before annotating

For me, this works only the first time. When I run `annotate` the second time, the comments are being removed.

How do I need to wrap or format the existing class documentation to protect it?

dben89x:
Yeah this is absolutely horrible. I lost so many comments that were not part of a commit. Why in the world would this not just prepend to a file?

PR #952, opened 2022-05-09
Option to add a blank line between annotation and class definition

cmer:
This adds an option to add a blank line separator between the annotation and the class definition of models. I believe this is important and should likely be the default in order to prevent deletion of comments above a class definition, as per issue #951.


cmer:
CI is broken -- failures are not related to the code in this PR.

Issue #969, opened 2022-10-04
Annotate gem does not support Rails autoloader collapsed directories


bazay:
Annotate gem does not support Rails autoloader collapsed directories i.e.

```ruby
pry(AnnotateModels)> Rails.autoloaders.main
=> #<Zeitwerk::Loader:0x00007f6305898b58
 @autoloaded_dirs=[],
 @autoloads={},
 @collapse_dirs=
  #<Set: {"/home/developer/my_app/app/components/my_namespace/models"}>,
 @collapse_glob_patterns=#<Set: {/home/developer/my_app/app/components/*/models", "/home/developer/my_app/app/components/my_namespace/jobs"}>,
 @eager_load_exclusions=#<Set: {}>,
 @eager_loaded=true,
 @ignored_glob_patterns=
  #<Set: {"/home/developer/my_app/lib/rails",
   "/home/developer/my_app/lib/patches",
   "//home/developer/my_app/lib/rails/generators/job",
...
pry(AnnotateModels)> Rails.autoloaders.main.collapsed_dirs
=> #<Set: {"/home/developer/my_app/app/components/my_namespace/models"}>
```
In `config/application.rb`:
```ruby
Rails.autoloaders.main.collapse("app/components/*/models")
```

I've obfuscated the model names and namespace, but the end result is still the same.

## Commands

```
$ bundle exec annotate
Unable to annotate app/components/my_namespace/models/my_model.rb: file doesn't contain a valid model class
Unable to annotate app/components/my_namespace/models/my_second_model.rb: file doesn't contain a valid model class
Unable to annotate app/components/my_namespace/models/my_third_model.rb: file doesn't contain a valid model class
$
```

## Version

- annotate (3.2.0 96831c1)
- rails (7.0.3.1)
- ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]

guigs:
Here's a monkey patch we are using that find classes based on zeitwerk loader. 
Our use case is for loading classes within namespaces. I'm not sure if it works with collapsed dirs, but maybe you can modify it to work.

Put in an initializer:

```ruby
# frozen_string_literal: true

if Rails.env.development?
  require 'annotate'

  module AnnotateModelsWithZeitwerk
    def get_model_class(file)
      loader = Rails.autoloaders.main
      root_dirs = loader.dirs(namespaces: true) # or `root_dirs = loader.root_dirs` with zeitwerk < 2.6.1
[...]

PR #970, opened 2022-10-04
[#969] Add support for Rails autoloader collapsed directories

bazay:
Add support for collapsed autoloader directories set using `Rails.autoloaders.main.collapse(...)`.

Checklist:
- [x] Add tests

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -611,6 +611,15 @@ module AnnotateModels
         $LOAD_PATH.map(&:to_s)
                   .select { |path| absolute_file.include?(path) }
                   .map { |path| absolute_file.sub(path, '').sub(/\.rb$/, '').sub(/^\//, '') }
+
+      # Handle Rails apps with collapsed model paths
+      model_paths = model_paths
+        .select do |model_path|
+          defined?(Rails) &&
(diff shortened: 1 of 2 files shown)


lib/annotate/annotate_models.rb
```diff
@@ -611,6 +611,15 @@ def get_loaded_model(model_path, file)
...
                   .select { |path| absolute_file.include?(path) }
                   .map { |path| absolute_file.sub(path, '').sub(/\.rb$/, '').sub(/^\//, '') }
+
+      # Handle Rails apps with collapsed model paths
+      model_paths = model_paths
+        .select do |model_path|
+          defined?(Rails) &&
+          Rails.autoloaders.main.collapse_dirs.any? &&
+          Rails.autoloaders.main.collapse_dirs.select { |path| path.match(model_path) }
+        end.map { |model_path| model_path.sub(/\/models/, '') }
```
bazay: My concern here is simply removing `"/models"` from the path is a hardcoded approach. To me, it could be great if we could use `ActiveSupport::Inflector` or `Zeitwerk` to define the expected path, as Zeitwerk enforces the behaviour set using `Rails.autoloaders.main.collapse(...)`.

PR #979, opened 2023-02-07
Support `--frozen` option for routing annotations

kg8m:
Hi. Thanks for this great gem! The `--frozen` option is very useful.

I noticed that the `--frozen` option doesn't work for routing annotations (`rake annotate_routes` and `annotate --routes`). This pull request adds the support for that.

Best regards.

### DIFF
--- a/lib/annotate/annotate_routes.rb
+++ b/lib/annotate/annotate_routes.rb
@@ -29,8 +29,7 @@ module AnnotateRoutes
         content, header_position = Helpers.strip_annotations(existing_text)
         new_content = annotate_routes(HeaderGenerator.generate(options), content, header_position, options)
         new_text = new_content.join("\n")
-
-        if rewrite_contents(existing_text, new_text)
+        if rewrite_contents(existing_text, new_text, options[:frozen])
           puts "#{routes_file} was annotated."
         else
           puts "#{routes_file} was not changed."
@@ -82,11 +81,11 @@ module AnnotateRoutes
       content
     end
 
-    def rewrite_contents(existing_text, new_text)
+    def rewrite_contents(existing_text, new_text, frozen)
       content_changed = (existing_text != new_text)
 
       if content_changed
-        abort "annotate error. #{routes_file} needs to be updated, but annotate was run with `--frozen`." if options[:frozen]        
+        abort "annotate error. #{routes_file} needs to be updated, but annotate was run with `--frozen`." if frozen        
         File.open(routes_file, 'wb') { |f| f.puts(new_text) }
       end
       
--- a/lib/annotate/annotate_routes.rb
+++ b/lib/annotate/annotate_routes.rb
@@ -30,13 +30,7 @@ module AnnotateRoutes
         new_content = annotate_routes(HeaderGenerator.generate(options), content, header_position, options)
         new_text = new_content.join("\n")
 
-        if options[:frozen]
-          if needs_rewrite_contents?(existing_text, new_text)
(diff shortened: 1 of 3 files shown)


ctran:
@kg8m Would you mind rebase against the latest `develop` (which has fixed the CI issues)?

Thanks.

lib/annotate/annotate_routes.rb
```diff
@@ -30,10 +30,18 @@ def do_annotations(options = {})
...
+          else
+            puts "#{routes_file} was not changed."
+          end
         else
-          puts "#{routes_file} was not changed."
+          if rewrite_contents(existing_text, new_text)
+            puts "#{routes_file} was annotated."
+          else
+            puts "#{routes_file} was not changed."
+          end
```
ctran: Rubocop wants to convert this into an `elsif`.

ctran:
Could I convince you to add an unit test/spec for this new option?

kg8m:
@ctran 
I fixed the RuboCop error and added tests for the `--frozen` option of routing annotations. Could you please review the changes?

Best regards.

lib/annotate/annotate_routes.rb
```diff
@@ -30,7 +30,13 @@ def do_annotations(options = {})
...
+        if options[:frozen]
+          if needs_rewrite_contents?(existing_text, new_text)
+            abort "annotate error. #{routes_file} needs to be updated, but annotate was run with `--frozen`."
+          else
+            puts "#{routes_file} was not changed."
+          end
+        elsif rewrite_contents(existing_text, new_text)
           puts "#{routes_file} was annotated."
         else
           puts "#{routes_file} was not changed."
```
ctran, line 42: I feel this can be somewhat improved to remove the code duplication and nested ifs.  What about something like this?

```ruby
# don't feel like a separate function is needed
content_changed = (existing_text.strip != new_text.strip)  

if !content_changed # you can flip this condition around if you want
  puts "#{routes_file} was not changed."
else
  abort "annotate error. #{routes_file} needs to be updated, but annotate was run with `--frozen`." if options[:frozen]
  rewrite_contents(existing_text, new_text)  
end
```

spec/lib/annotate/annotate_routes_spec.rb
ctran, line 1: Thanks for the test!

ctran (changes requested):

lib/annotate/annotate_routes.rb
ctran: Now that I'm looking at the code closer, maybe we should move all this inside the `rewrite_contents` function.

ctran, line 35: This could be improved to remove code duplication and nested ifs:

```ruby
if rewrite_contents(existing_text, new_text)
  puts "#{routes_file} was annotated."
else
  puts "#{routes_file} was not changed."
end
```

Then change `#rewrite_contents` to:
```ruby
def rewrite_contents(existing_text, new_text)
  content_changed = (existing_text != new_text)

  # Rubocop may prefer "abort ... if"
  if content_changed && options[:frozen] 
    abort "annotate error. #{routes_file} needs to be updated, but annotate was run with `--frozen`." 
  end

  File.open(routes_file, 'wb') { |f| f.puts(new_text) }
  content_changed
end
```

ctran (approved):
I made the suggestions directly on your fork, hope you don't mind.

kg8m:
@ctran 
Thank you so much!

PR #980, opened 2023-03-28
Fix flaky specs

lovro-bikic:
Fixes flaky specs which fail the CI.

I ran the test suite a couple of hundred of times locally with different seeds to make sure nothing else fails. I couldn't find any other issues.

### DIFF
--- a/spec/lib/annotate/annotate_models_spec.rb
+++ b/spec/lib/annotate/annotate_models_spec.rb
@@ -153,7 +153,11 @@ describe AnnotateModels do
     end
 
     before :each do
-      AnnotateModels.send(:parse_options, options)
+      AnnotateModels.parse_options(options)
+    end
+
+    after :each do
+      AnnotateModels.parse_options({ skip_subdirectory_model_load: false })
     end
 
     describe '@root_dir' do
@@ -1647,6 +1651,10 @@ describe AnnotateModels do
       Annotate::Helpers.true?(ENV['show_complete_foreign_keys'])
     end
 
+    after :each do
+      ENV.delete('show_complete_foreign_keys')
+    end
+
     context 'when default value of "show_complete_foreign_keys" is not set' do
       it 'returns false' do
         is_expected.to be(false)
@@ -1658,13 +1666,13 @@ describe AnnotateModels do
         Annotate.set_defaults('show_complete_foreign_keys' => 'true')
       end
 
+      after do
+        Annotate.instance_variable_set('@has_set_defaults', false)
+      end
+
       it 'returns true' do
         is_expected.to be(true)
       end
-    end
-
-    after :each do
-      ENV.delete('show_complete_foreign_keys')
     end
   end
 
@@ -1815,8 +1823,14 @@ describe AnnotateModels do
   end
 
   describe '.get_model_class' do
-    before :all do
-      AnnotateModels.model_dir = Dir.mktmpdir('annotate_models')
+    before :each do
+      @model_dir = Dir.mktmpdir('annotate_models')
+      AnnotateModels.model_dir = @model_dir
+      create(filename, file_content)
+    end
+
+    after :each do
+      FileUtils.remove_dir(@model_dir, true)
     end
 
     # TODO: use 'files' gem instead
@@ -1827,10 +1841,6 @@ describe AnnotateModels do
           f.puts(file_content)
         end
       end
-    end
-
-    before :each do
-      create(filename, file_content)
     end
 
     let :klass do
@@ -2113,7 +2123,9 @@ describe AnnotateModels do
 
         let :file_content_2 do
           <<-EOS
-            class Bar::Foo < ActiveRecord::Base
+            module Bar
+              class Foo < ActiveRecord::Base
+              end
             end
           EOS
         end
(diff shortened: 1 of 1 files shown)


spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -153,7 +153,11 @@ def mock_column(name, type, options = {})
     end
 
     before :each do
-      AnnotateModels.send(:parse_options, options)
+      AnnotateModels.parse_options(options)
+    end
+
+    after :each do
+      AnnotateModels.parse_options({ skip_subdirectory_model_load: false })
```
lovro-bikic, line 160: One of the tests sets `skip_subdirectory_model_load` to `true` and this causes issues for other tests which expect the value to be the default false. This fix resets the value to the default.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -1658,14 +1666,14 @@ def mock_column(name, type, options = {})
         Annotate.set_defaults('show_complete_foreign_keys' => 'true')
       end
 
+      after do
+        Annotate.instance_variable_set('@has_set_defaults', false)
+      end
```
lovro-bikic, line 1671: `Annotate.set_defaults` is used in a couple of tests, but once called it will set `@has_set_defaults` to true which will prevent the method from being called a second time. This fix ensures that the method can be called in other tests by setting the instance variable back to false.

Note that this line was already added in another place in this file for the same purpose, so it was probably just missed here.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -1815,8 +1823,14 @@ def mock_column(name, type, options = {})
...
-      AnnotateModels.model_dir = Dir.mktmpdir('annotate_models')
+    before :each do
+      @model_dir = Dir.mktmpdir('annotate_models')
+      AnnotateModels.model_dir = @model_dir
+      create(filename, file_content)
+    end
+
+    after :each do
+      FileUtils.remove_dir(@model_dir, true)
     end
```
lovro-bikic, line 1834: Refactored so that a different directory is used for each test and cleaned up afterwards.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -2113,7 +2123,9 @@ class Foo < ActiveRecord::Base
 
         let :file_content_2 do
           <<-EOS
-            class Bar::Foo < ActiveRecord::Base
+            module Bar
+              class Foo < ActiveRecord::Base
+              end
```
lovro-bikic, line 2128: `class Bar::Foo` only works if module `Bar` already exists (source https://www.akshaykhot.com/ruby-difference-nested-modules-syntax/), but this is not the case for this spec.

The reason it sometimes worked before is because another test created the `Bar` module, so this one then wouldn't fail because it could use it.

With this fix module Bar will be created if it doesn't exist already, so this test doesn't depend on other tests.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -2163,6 +2177,7 @@ class Bar::Foo < ActiveRecord::Base
         it 'attempts to load the model path without expanding if skip_subdirectory_model_load is false' do
           allow(AnnotateModels).to receive(:skip_subdirectory_model_load).and_return(false)
           full_path = File.join(AnnotateModels.model_dir[0], filename_2)
+          Kernel.load(full_path)
```
lovro-bikic, line 2180: This spec seems to expect `full_path` to be loaded as part of the setup, but it was missing before. There are other similar tests which have the `Kernel.load` step as part of the setup, so I reused it here.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -2519,6 +2535,10 @@ class User < ActiveRecord::Base
       Annotate::Helpers.reset_options(Annotate::Constants::ALL_ANNOTATE_OPTIONS)
     end
 
+    after :each do
+      FileUtils.remove_dir(@model_dir, true)
+    end
```
lovro-bikic, line 2544: Added clean-up of directory after each test.

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -2531,7 +2551,7 @@ def annotate_one_file(options = {})
       Annotate.set_defaults(options)
       options = Annotate.setup_options(options)
       AnnotateModels.annotate_one_file(@model_file_name, @schema_info, :position_in_class, options)
-
+    ensure
```
lovro-bikic, line 2558: I added the ensure here because the lines below are all part of clean-up, but they sometimes wouldn't get called because `AnnotateModels.annotate_one_file` can raise an error (this is expected behavior in certain cases and is part of the tests, but because the error is raised clean-up isn't performed so the settings end up in unexpected state for other tests).

spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -2218,6 +2234,10 @@ class Foo < ActiveRecord::Base
       AnnotateModels.remove_annotation_of_file(path)
     end
 
+    after :each do
+      FileUtils.remove_dir(tmpdir, true)
+    end
```
lovro-bikic, line 2239: Added clean-up of tmp directory.

ctran:
Thanks!!!

ctran (approved):
Much appreciated!

PR #986, opened 2023-05-16
Correctly capture columns that have a multibyte comment

### LINKED ISSUE #836: Comments by multi-byte characters are ignored
Hi, everyone. So thank you for great product.

I occurred in trouble and want to hear everyone's opinion.

I'm Japanese developer so often add some comment for each database column in Japanese.

For example:

```ruby
create_table "users", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci", comment: "アユーザ", force: :cascade do |t|
  t.string "name", null: false, comment: "名前"
  t.string "email", null: false, comment: "メールアドレス"
end
```

However, they're ignored because they don't match a regular expression which checks diff between old schema and new schema, It's specified [here](https://github.com/ctran/annotate_models/blob/27e5965260cfd60d51054bb82aeef3b47ea22c46/lib/annotate/annotate_models.rb#L378).

If we don't have any problems, I want to fix the problem as follows.

```ruby
# current
column_pattern = /^#[\t ]+[\w\*\.`]+[\t ]+.+$/

# I expect
column_pattern = /^#[\t ]+[^\t ]+[\t ]+.+$/
```

What do you think?

## Commands

```
$ bundle exec annotate
Model files unchanged.
```

## Version

- annotate version
  - 3.1.1
- rails version
  - 6.0.3.2
- ruby version
  - 2.7.1

y-yagi:
Fixes #836.

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -433,7 +433,7 @@ module AnnotateModels
       old_header = old_content.match(header_pattern).to_s
       new_header = info_block.match(header_pattern).to_s
 
-      column_pattern = /^#[\t ]+[\w\*\.`]+[\t ]+.+$/
+      column_pattern = /^#[\t ]+[^\t ]+[\t ]+.+$/
       old_columns = old_header && old_header.scan(column_pattern).sort
       new_columns = new_header && new_header.scan(column_pattern).sort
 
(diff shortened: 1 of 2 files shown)


y-yagi:
I've rebased with the latest code.

y-yagi:
I've rebased with the latest code again.

drwl:
@y-yagi I added your changes to my gem if you need the multibyte comment support https://github.com/drwl/annotaterb

PR #992, opened 2023-07-25
Updated options documentation in README

matheus-poli:
Options documentation was out of date. I ran `annotate --help` on the latest version of the gem and updated the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,24 +197,25 @@ you can do so with a simple environment variable, instead of editing the
 ## Options
 
     Usage: annotate [options] [model_file]*
-            --additional-file-patterns   Additional file paths or globs to annotate, separated by commas (e.g. `/foo/bar/%model_name%/*.rb,/baz/%model_name%.rb`)
+            --additional-file-patterns path1,path2,path3
+                                        Additional file paths or globs to annotate, separated by commas (e.g. `/foo/bar/%model_name%/*.rb,/baz/%model_name%.rb`)
         -d, --delete                     Remove annotations from all model files or the routes.rb file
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -236,21 +236,20 @@ you can do so with a simple environment variable, instead of editing the
...
-        -f [bare|rdoc|yard|markdown],    Render Schema Infomation as plain/RDoc/YARD/Markdown
+        -f [bare|rdoc|yard|markdown],    Render Schema Infomation as plain/RDoc/Yard/Markdown
             --format
             --force                      Force new annotations even if there are no changes.
             --frozen                     Do not allow to change annotations. Exits non-zero if there are going to be changes to files.
             --timestamp                  Include timestamp in (routes) annotation
             --trace                      If unable to annotate a file, print the full stack trace, not just the exception message.
-        -I, --ignore-columns REGEX       don't annotate columns that match a given REGEX (e.g. `annotate -I '^(id|updated_at|created_at)'`)
-            --ignore-routes REGEX        don't annotate routes that match a given REGEX (e.g. `annotate -I '(mobile|resque|pghero)'`)_
+        -I, --ignore-columns REGEX       don't annotate columns that match a given REGEX (i.e., `annotate -I '^(id|updated_at|created_at)'`
```
BryanH, line 245: The original is correct here; these are examples. 

* i.e. (latin: _id est_), means "in other words" 
* e.g. (latin: _exemplī grātiā_), means "for example"

If we substitute the meaning for the abbreviation, it becomes clearer which one fits.

> ... don't annotate columns that match a given REGEX (in other words, `annotate -I '^...`

> ... don't annotate columns that match a given REGEX (for example, `annotate -I '^...`

BryanH (changes requested):
Please undo the _e.g._ ->  _i.e._ updates and it should be good.

PR #994, opened 2023-08-03
Add support for factory_bot's default suffixed pattern

### LINKED ISSUE #993: Does not find factories with default suffix example
The default suffixed generator example in https://github.com/thoughtbot/factory_bot_rails states:

```ruby
config.generators do |g|
  g.factory_bot suffix: "factory"
end
```

This configuration will result in factories being created with this pattern:

`spec/factories/<pluralized_model_name>_factory.rb` and we only have support for pluralized model name without suffix.

## Version

- annotate version: 3.2.0
- rails version: 7.0.6
- ruby version: 3.2.2

meric426:
Fixes #993


stefanahman (approved):
Thank you!

PR #999, opened 2023-11-15
Make `with_comment_column` work with `Annotate.set_defaults`

### LINKED ISSUE #990: Make `with_comment_column` work with `Annotate.set_defaults`
Describe your problem here.

## Commands

```
$ rake annotate_models
```

## Acceptance Criteria

- `"with_comment_column" => "false"` appears after `"with_comment" => "false"` when calling `rails generate annotate:install`.
- its setting is read when annotating (right now, setting it to `true` does not activate the switch and comment column is not rendered).

## Related Issues

This was an oversight when implementing #988.

Adeynack:
Fixes #990

### DIFF
--- a/spec/lib/tasks/annotate_models_spec.rb
+++ b/spec/lib/tasks/annotate_models_spec.rb
@@ -1,0 +1,30 @@
+require_relative '../../spec_helper'
+
+describe 'annotate_models rake task and Annotate.set_defaults' do
+  before do
+    Rake.application = Rake::Application.new
+    Rake::Task.define_task('environment')
+    Rake.load_rakefile('tasks/annotate_models.rake')
+  end
+
+  let(:annotate_models_argument) do
+    argument = nil
+    expect(AnnotateModels).to receive(:do_annotations) { |arg| argument = arg }
+    Rake::Task['annotate_models'].invoke
+    argument
+  end
+
+  describe 'with_comment_column' do
+    after { ENV.delete('with_comment_column') }
+    subject { annotate_models_argument[:with_comment_column] }
+
+    context 'when Annotate.set_defaults is not called (defaults)' do
+      it { is_expected.to be_falsey }
+    end
+
+    context 'when Annotate.set_defaults sets it to "true"' do
+      before { Annotate.set_defaults('with_comment_column' => 'true') }
+      it { is_expected.to be_truthy }
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


sureshc:
Thank you! I'm not familiar with this codebase. Are there any tests that need to be updated to validate this change?

Adeynack:
@sureshc : The only test I found that seems to concern `.set_defaults` are strangely specific to ONE configuration key, so I decided to not touch it. @ctran can maybe point me at where it would be decent to test this thing. I am also very uncomfortable in submitting modifications without any kind of automated testing proving its worth.

ctran:
> @sureshc : The only test I found that seems to concern `.set_defaults` are strangely specific to ONE configuration key, so I decided to not touch it. @ctran can maybe point me at where it would be decent to test this thing. I am also very uncomfortable in submitting modifications without any kind of automated testing proving its worth.

Maybe add a new test in https://github.com/ctran/annotate_models/blob/master/spec/lib/tasks/annotate_models_migrate_spec.rb?

Adeynack:
@ctran : I added a brand new spec file for it, since I judged it was a separate concern than what's in `annotate_models_migrate_spec`. But thanks a lot for this hint, it pushed me directly in the right direction! 🤩

spec/lib/tasks/annotate_models_spec.rb
```diff
@@ -0,0 +1,32 @@
+require_relative '../../spec_helper'
+
+describe 'Annotate annotate_models rake task and Annotate.set_defaults' do # rubocop:disable RSpec/DescribeClass
```
Adeynack, line 3: @ctran : My apologies for bypassing a `rubocop` rule here, but there is no class to describe Rake tasks, so that would justify the exception.

Also, I do not understand why `rubocop` is not complaining about `spec/lib/tasks/annotate_models_migrate_spec.rb`, since it also has a free string test root describe block.

Adeynack:
@sureshc @ctran : Code is complete & tests were written. Please review 😺

spec/lib/tasks/annotate_models_spec.rb
ctran: It was excluded: https://github.com/ctran/annotate_models/blob/develop/.rubocop_todo.yml

ctran (approved):
LGTM - Thanks!

spec/lib/tasks/annotate_models_spec.rb
Adeynack: Oh. Then it probably would have been better that I add the exception of my file the same way.  But it's also not a big deal and it's already merged. Thanks!

PR #1000, opened 2023-12-03
Add parentheses in allowed column pattern for commented columns

m11o:
This PR adds parentheses in allowed column pattern for commented columns.

The annotation of a model with commented columns is not updated even if the column will be updated.

### Steps to reproduce

1. Create a new Rails application and use MySQL or PostgreSQL database
2. Add annotate_models gem in Gemfile
3. Create a model with a commented column
    - e.g.
    - ```ruby
      create_table :users do |t|
        t.string :name, comment: 'first_name + last_name'
        t.string :email, null: false

        t.timestamps
      end
      ```
5. Change the commented column from default null to non-null.
    - e.g.
    - ```ruby
      change_column_null :users, :name, false
      ```
7. The annotation of the model isn't updated
    - ```
      Model files unchanged.
      ```
    - but, schema.rb is updated.
    - ```diff
      -ActiveRecord::Schema[7.0].define(version: 2023_12_03_034600) do
      +ActiveRecord::Schema[7.0].define(version: 2023_12_03_040324) do
         create_table "users", charset: "utf8mb4", collation: "utf8mb4_0900_ai_ci", force: :cascade do |t|
      -    t.string "name", comment: "first_name + last_name"
      +    t.string "name", null: false, comment: "first_name + last_name"
           t.string "email", null: false
           t.datetime "created_at", null: false
           t.datetime "updated_at", null: false
      ```

### DIFF
--- a/spec/lib/annotate/annotate_models_spec.rb
+++ b/spec/lib/annotate/annotate_models_spec.rb
@@ -79,7 +79,8 @@ describe AnnotateModels do
       limit: nil,
       null: false,
       default: nil,
-      sql_type: type
+      sql_type: type,
+      comment: nil
     }
 
     stubs = default_options.dup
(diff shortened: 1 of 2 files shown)


spec/lib/annotate/annotate_models_spec.rb
```diff
@@ -3048,6 +3049,58 @@ class User < ActiveRecord::Base
...
+              schema_info,
+              :position_in_class,
+              @annotation_options
+            )
+          end
+            .to(
+              change { File.read(@model_file_name) }
+                .from(include("name(commented column) :string\n"))
+                .to(include("name(commented column) :string           not null\n"))
+            )
```
m11o, line 3099: If parentheses aren't added, this annotation isn't updated and this test is failed with the following error.

```
  1) AnnotateModels annotating a file if a table has a commented column and the model already has annotation and the commented column is updated to non-null should be added non-null option in the commented column
     Failure/Error:
       expect do
         AnnotateModels.annotate_one_file(@model_file_name, schema_info, :position_in_class, @annotation_options)
       end.to change { File.read(@model_file_name) }
                .from(include("name(commented column) :string\n"))
                .to(include("name(commented column) :string           not null\n"))

       expected `File.read(@model_file_name)` to have changed from include "name(commented column) :string\n" to include "name(commented column) :string           not null\n", but did not change
     # ./spec/lib/annotate/annotate_models_spec.rb:3087:in `block (5 levels) in <top (required)>'
```

BryanH (approved):

PR #1036, opened 2024-12-26
fix: work with rails 8

wlads:
- **Runs annotate on 'rake db:rollback' too**
- **Revert "Runs annotate on 'rake db:rollback' too"**
- **bump: version to 3.2.0**
- **Bump version to 3.2.0**
- **fix: work with rails 8**


wlads:
Changes on top of main (v3.2.0)

Issue #1039, opened 2025-01-25
‼️ THIS GEM IS DEPRECATED. SWITCH TO https://github.com/drwl/annotaterb ‼️


rafaelsales:
The maintained fork is at https://github.com/drwl/annotaterb

For fun:
![Image](https://github.com/user-attachments/assets/ab5382d8-7554-4095-aec0-e9e096a574d9)

brupla6126:
Would be a great idea to put this message at the top of README.MD

corsonknowles:
> Would be a great idea to put this message at the top of README.MD

If only there was a maintainer who could do that!

PR #1040, opened 2025-02-05
Add deprecation notice and link to annotaterb to the README

sethherr:
This gem is no longer maintained. It doesn't support Rails 8.

This PR adds a note that the gem is no longer maintained and recommends looking at [annotaterb](https://github.com/drwl/annotaterb) for an alternative.

See e.g. https://github.com/ctran/annotate_models/issues/1039

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # This Gem is no longer maintained
 
-**Consider [annotate_rb](https://github.com/drwl/annotaterb)** for an actively maintained alternative.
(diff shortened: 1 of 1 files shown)


rafaelsales:
@ctran

ApaeP (approved):
This seems like a useful and necessary update.
Clarifying the deprecation status helps avoid confusion. 🙂

koliao (approved):

PR #1042, opened 2025-05-07
Account for object default values when annotating models

mochetts:
Fix Default Value Handling for Objects with to_s Method

This PR updates the quote method in annotate_models.rb to properly handle default values that are objects implementing the `to_s` method. This change improves the annotation behavior when dealing with complex default values in model attributes.

Changes:
- Updated the quote method to properly handle object default values
- Added tests to verify the new behavior

Previous behavior:
- Objects with `to_s` implementation weren't properly quoted in annotations

New behavior:
- Objects with `to_s` implementation are now properly quoted in the annotations

This change ensures more accurate model annotations when dealing with complex default values.

With this change we can effectively annotate object defaults. This is an example with the woking fix in one of our repos:

Screenshot 2025-05-07 at 9 04 59 AM

### DIFF
--- a/lib/annotate/annotate_models.rb
+++ b/lib/annotate/annotate_models.rb
@@ -107,7 +107,7 @@ module AnnotateModels
       when BigDecimal               then value.to_s('F')
       when Array                    then value.map { |v| quote(v) }
       else
-        value.inspect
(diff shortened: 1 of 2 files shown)


mochetts:
This gem is not maintained anymore

PR #1048, opened 2026-01-13
Update gemspec to allow Rails 8

DSalko:


lucascaton:
cc/ @ctran

braindeaf:
@ctran or this? https://github.com/ctran/annotate_models/pull/1052

PR #1053, opened 2026-09-16
Support Rails 8 and Ruby 4

vineelvineel:
Adds support for Rails 8 and Ruby 4, without dropping anything older.

- Widens the `activerecord` constraint to `< 9.0`. The `< 8.0` cap makes bundler silently resolve `annotate` down to 2.6.5 (2014) on a Rails 8 app instead of reporting a conflict.
- Restores `File.exists?` for the test-only `files` gem, and accepts Ruby 4's backtrace quoting in two specs, so the suite runs on Ruby 4. (`File.exists?` already fails on Ruby 3.3, printing `NoMethodError` at teardown on every run today.)
- Drops the `ruby_dep '1.5.0'` pin, which caps Ruby below 3.x.
- Adds a `gemfiles/` entry per activerecord minor from 6.1 to 8.1, with a CI leg for each, so the supported range is tested rather than asserted.

No library code needed changing: `klass.connection`, the only ActiveRecord API in `annotate_models.rb` I expected to be a problem, is still present in 8.0 and 8.1 and merely marked *"Soft deprecated. Use `#with_connection` or `#lease_connection` instead."*

### Verified locally

Every combination below ran the full suite — **392 examples, 0 failures**:

| ruby | activerecord |
| --- | --- |
| 3.1.6 | 6.1.7.10 |
| 3.3.4 | 6.1.7.10, 7.0.10, 7.1.6, 7.2.3.2, 8.0.5.1, 8.1.3.1 |
| 3.4.9 | 7.2.3.2, 8.0.5.1, 8.1.3.1 |
| 4.0.7 | 7.2.3.2, 8.0.5.1, 8.1.3.1 |

`gemfiles/rails_6.1.gemfile` pins `concurrent-ruby < 1.3.5`: activesupport below 7.1 relies on concurrent-ruby requiring `logger`, which it stopped doing in 1.3.5. Without the pin, activesupport 6.1 raises `uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger` on load — unrelated to this gem, but it is what keeps a 6.1 leg from running at all.

The default `Gemfile` and its Ruby 2.7 leg are untouched, and rubocop still runs only there.


vineelvineel:
@ctran As discussed with you yesterday, please find the PR for Rails 8 and Ruby 4 support. Please let me know if you have any questions. Thanks

